Issue #1, opened 2020-12-30
NPCs wake up immediately 


AmProsius:
NPCs don't sleep through and wake up immediately when they have gone to bed.

catalinstoian:
`/************************************************
*              NSC benutzt Bett,                *
* wenn kein Bett da, macht der ZS_StandAround   *
*************************************************/

func void ZS_Sleep()
{
	PrintDebugNpc 		(PD_TA_FRAME,"ZS_Sleep");
	
	B_SetPerception		(self);    
	AI_SetWalkmode 		(self,	NPC_WALK);
	if (!C_BodyStateContains(self,BS_MOBINTERACT))
	{
		if (Hlp_StrCmp(Npc_GetNearestWP(self), self.wp)== 0)
        {
			AI_GotoWP	(self, self.wp);
		};
	};
	AI_StartState		(self, ZS_SleepBed, 1, "");
};		
		
func void ZS_SleepBed()		
{
    PrintDebugNpc 		(PD_TA_FRAME,"ZS_SleepBed_Loop");
    
    LightSleepPerception();
    
	if (Wld_IsMobAvailable (self,"BED"))
	{
	    AI_UseMob		(self,	"BED",1);
	}
	else if (Wld_IsMobAvailable (self,"BEDHIGH")||Wld_IsMobAvailable (self,"BEDLOW"))
	{
	    AI_UseMob		(self,	"BEDHIGH",1); //Mehrfache Mob-Benutzung okay, weil nachfolgende UseMobs ignoriert werden, wenn UseMob schon aktiv ist
	    AI_UseMob 		(self,	"BEDLOW",1); // Mobs werden nur benutzt, wenn NSC sie sehen kann und wenn sie frei sind (sagt Ulf)
	}
	else
	{
		AI_StartState	(self, ZS_StandAround, 1, "");
	};
};

func int ZS_SleepBed_Loop()
{
	PrintDebugNpc 		(PD_TA_LOOP,"ZS_SleepBed_Loop");

	// SN: Workaround dafür, daß Wld_IsMobAvailable() zwar TRUE geliefert hat, AI_UseMob() aber gescheitert ist!
	if	(( !Npc_GetBodyState(self) == BS_MOBINTERACT	||	!Npc_GetBodyState(self) == BS_MOBINTERACT_INTERRUPT	)
	&&	(Npc_GetStateTime(self) > 3))
	{
		AI_StartState	(self, ZS_SitAround, 1, "");
[...]

PR #67, opened 2021-01-16
#1 make NPCs sleep properly

szapp:
This PR is not ready for merging yet. The issue below should be addressed first.

Are you sure this works as expected? A check against `Npc_GetBodyState` rarely renders true, because the body state is a bitfield, a combination of multiple active body states. That's why there is the function `C_BodyStateContains`. Is there a particular reason why you used `Npc_GetBodyState`?

_Originally posted by @szapp in https://github.com/AmProsius/gothic-1-community-patch/issues/1#issuecomment-761538175_

To test this fix, manual confirmation is required: When running `test 1` from the console, the hero is teleported to a hut at night. The user can confirm if the NPC inside remains asleep.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix001_FixNpcSleep.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix001_FixNpcSleep.d
@@ -1,0 +1,62 @@
+/*
+ * #1 NPCs wake up immediately
+ */
+func void Ninja_G1CP_001_FixNpcSleep() {
+    HookDaedalusFuncS("ZS_SleepBed_Loop", "Ninja_G1CP_001_FixNpcSleep_Hook");
+};
+
+/*
+ * This function intercepts the NPC state to introduce more conditions
+ */
+func int Ninja_G1CP_001_FixNpcSleep_Hook() {
+    // Changing the condition to start the ZS_SitAround is not easily possible with completely overwriting the function
+    // Overwriting the function should be avoided, because that would remove any additional changes from someone else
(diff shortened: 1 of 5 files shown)


szapp:
Ready for merge

AmProsius:
@szapp can you resolve the conflicts? I'm not sure about the contents of `misc.d`.

AmProsius:
I'm not sure if that's the intended behavior: When I run `test 1`, I'm teleported in front of a hut, the NPC inside is awake in the middle of the hut and immediately goes to bed. If I run `test 1` again, the NPC is resetted to the middle of the hut and goes to bed again.

szapp:
Yes, this seems correct. The test is a bit "light" and depends pretty much on observation. The thing to look out for is whether the NPC stays asleep or immediately gets up out of bed again and stands around. There are no prints or anything automatic to double check that.

You can remove the fix and run the test again to compare. To do so, comment out/remove this line:

https://github.com/AmProsius/gothic-1-community-patch/blob/9e7e735f2d01a38216edd5f4a78f21b39b939171/src/Ninja/G1CP/Content/NinjaInit.d#L17

AmProsius:
Thanks, I will try that.

AmProsius:
I can now confirm that this fix works as intended. Without the fix, the NPC gets out of bed and ith the fix the NPC remains sleeping. Great work!

szapp:
Thanks for double checking. That was the first test with "manual confirmation" that I wrote, that's why the automation and instruction was lacking a bit.

PR #69, opened 2021-01-17
Add pre-menu splash screen

szapp:
Let's discuss first, whether this is something we want.

GothicMod-2021-01-17-17-17-41-20


AmProsius:
Dropped this feature for the more subtle version indicator.

PR #72, opened 2021-01-19
#2 NPC use doors properly

szapp:
NPCs now use doors properly.

This PR should be ready to merge. An interesting thing to check would be, whether it also works when using with mods that already have fixed the issue.

The test involves manual confirmation with `test 2`.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix002_FixDoor.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix002_FixDoor.d
@@ -19,6 +19,7 @@ func void Ninja_G1CP_002_FixDoor_Hook() {
         && (Ninja_G1CP_GetAIVar(Hlp_GetInstanceID(self), "AIV_MOVINGMOB", 1)) { // AI-var was set to 1 or does not exist
             AI_UseMob(self, door,  1);                                          // Just to be safe
             AI_UseMob(self, door, -1);                                          // Actual fix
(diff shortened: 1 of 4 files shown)


AmProsius:
> An interesting thing to check would be, whether it also works when using with mods that already have fixed the issue.
Good point. I will test the fixes with the UnPatch, because that's the mod with almost all fixes we did.

AmProsius:
@szapp I've tested this fix with the main game and the UnPatch mod under the same conditions (Ninja + Union + 4 GB Patch).

Main game: The guard opens the door but then stops and doesn't move any further. If I start the test again (and the door is open), he walks through the open door.

UnPatch mod: The guard opens the door and walks through it. I guess this is the intended behavior.

There seems to be something that the UnPatch mod addresses and we currently don't.

szapp:
Thanks for testing?

> When comparing the code, `AI_ContinueRoutine (self);` seems to be missing.

That could be it. I wanted to provide the most minimal fix possible so I left it out (because it seemed to work without). We should add it and try again.

szapp:
I looked into it, and `AI_ContinueRoutine` is actually just treating the symptoms. I will push a change that replaces the call to `AI_ContinueRoutine` with removing the underlying issue instead. Would you mind testing with it again? I will do my best to test, too.

AmProsius:
> Would you mind testing with it again? I will do my best to test, too.

Of course, no problem. Thank you for looking into this again!

PR #74, opened 2021-01-20
#9: Remove freeze from fleeing

szapp:
This PR should be ready.

Testing this fix is a bit tricky, because the NPC has to be visible all times while fleeing from the player. The rest requires to position the hero in a location surrounded by as many NPCs as possible to increase the chances of observing (not observing) the freeze. It's best to run the test first without and then with the fix, to remember where freezes do happen without the fix. To do so remove the fix here: 

https://github.com/AmProsius/gothic-1-community-patch/blob/ca8adc9d50a1c8e48340cbfd70702ccdddde4f72/src/Ninja/G1CP/Content/NinjaInit.d#L20

Run the test with `test 9`.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix009_FixFlee.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix009_FixFlee.d
@@ -1,0 +1,36 @@
+/*
+ * #9 NPCs freeze when fleeing
+ */
+func void Ninja_G1CP_009_FixFlee() {
+    HookDaedalusFuncS("ZS_Flee_Loop", "Ninja_G1CP_009_FixFlee_Hook");
+};
+
+/*
+ * This function intercepts the NPC state to introduce more conditions
+ */
+func int Ninja_G1CP_009_FixFlee_Hook() {
+    // Temporarily disable AI_Wait
+    const int AI_Wait_popped = 6644536; //0x656338
+    const int once = 0;
+    if (!once) {
+        MemoryProtectionOverride(AI_Wait_popped, 4);
+        once = 1;
+    };
+    const int orig = 3296983179; /*8B F8 83 C4*/
+    const int newb = 3296984881; /*31 FF 83 C4*/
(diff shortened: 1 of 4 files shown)


AmProsius:
So this one was hard to test, but here's what I've experienced:
1. Main game (no fix, no Union): I teleported to the temple court and noticed several freezes. So the issue definitely existed.
2. Main game (no fix, Union): I teleported again and didn't notice any freeze.
3. Main game (fix, Union): Same as 2.
4. UnPatch (no fix, Union): Same as 2.

So I tested more locations for test case 2, for example New Camp and Old Camp. I didn't notice any freeze even without applying the fix. I think it might be fixed in Systempack/Union. I read the Union fixes on their [page](https://worldofplayers.ru/threads/40376/), but couldn't find a fix for this special situation (though it might be fixed with something like `No_Interrupt_Interact_ANI`. I propose we close this PR for now until someone can confirm the issue still exists.

szapp:
It would be tricky for Union/SystemPack to fix this issue, because it is not an engine bug, but a scripting bug. But who knows maybe it's somehow fixed. Either way, I would vote for keeping this change, because it is doing something very minor:

The reason for the freeze is that after reaching a position to flee to and before determining the next position to run towards, the NPC is told to wait for one second. At best this causes the NPC to stand still, at worst it causes a freeze in the animation. All the fix here does is intercept any calls to the "wait" function made during the flee state and aborts them.

AmProsius:
I understand. As this PR doesn't do anything wrong, I merged it to master.

szapp:
Thanks, I will test it, too. Just to ease our minds

szapp:
To me it seemed like the issue also occurred with Union (without the fix). I tried it from the waypoint "PSI_TEMPLE_NOVIZE_PR4". One of the novices would always run up the small podium (from which the baal preaches) and wait/freeze up there for one second. I will push another change for increased compatibility (concerns only the test, not the fix).

PR #87, opened 2021-01-27
#53: Make ladders in old mine usable

### LINKED ISSUE #53: Ladder in Old Mine too short
The ladder that leads back up from the cave with the cog is too short. It is hard to reach the top without falling down again.

szapp:
### Description
Fixes #53. This change elongates and re-positions the two ladders in the old mine to stop the player from falling down. Since the ladders are changed to be vertical, the gap to the cliff is bridged with a wooden plank in the style of the old mine (same visual).

### Test
Run manual test with `test 53`.

#### Details 
As for any game save sensitive fix, testing will have to be a bit more elaborate:
1. Run the test on a new game
2. Run the test on a loaded game (prior to the fix, or preferably from vanilla Gothic or a mod)
3. Start the game with the fix active
    - Run the test
    - Save the game
    - Unload the fix (or preferably the entire patch, leaving vanilla Gothic or mod)
    - Load the game save
    - Confirm the fix was reverted

It should be briefly visible that the fix is reverted when saving.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Gamesave/fix053_OldMineLadders.d
+++ b/src/Ninja/G1CP/Content/Fixes/Gamesave/fix053_OldMineLadders.d
@@ -1,0 +1,186 @@
+/*
+ * #53 Ladder in Old Mine too short
+ */
+
+/* Old positions of the ladders (Integerfloats for precision) */
+const int Ninja_G1CP_053_OldMineLadders_PosL1old[16] = {1056645348, 1039739221,  1063053323, -996629510,
+                                                        1036063365, 1064967154, -1102844970, -975445247,
+                                                       -1084372949, 1043555876,  1055881796, -981167673,
+                                                                 0,          0,           0, 1065353216};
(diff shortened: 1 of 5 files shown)


AmProsius:
So here is what my tests showed:
* The fix works as intended on both new and loaded game saves.
* The fix is reverted when starting or loading a game without the patch.
* When starting or loading a game with just Union (meening Systempack/Union), the hero doesn't fall down of the ladders in their original position.
* When starting or loading the vanilla game, the issue is still present.

It looks to me like Union already fixes this bug somehow. The question is, do we still fix this issue too or do we simply require Union? @szapp can you confirm the behavior of the ladders without the patch but with Union?

I admire your fix, but I think I'd like to close this PR if it turns out that the issue is already fixed. Do you agree?

szapp:
Thank you for testing!

Yes it seems like it works with either SystemPack or Union. I don't know how reliable, but I tried it a handful of times (both with and without). I agree, go ahead and close the PR. Since we still have it here, we could always go back if we change our minds.

Maybe we should double check any mechanical bugs with SystemPack and Union before implementing them. Is this what you had in mind with the label "verification needed?". Might be worth adding that label again (I think it's gone).

PR #88, opened 2021-01-28
#50: Fix direction and collision of pillar

### LINKED ISSUE #50: Column in monastery ruin falls in wrong direction
The column in the monastery ruin falls in the wrong direction.

szapp:
### Description
Fixes #50. The falling direction and the collision of the pillar is corrected. The hero can now properly cross over the cliff at the monastery ruins.

### Test
Run manual test with `test 50`. The player is teleported to the pillar (provided they are in the correct world).

Here, it would be interesting to see if the fix is in conflict with other fixes and is properly reverted when removing the patch.


AmProsius:
Here's what I noticed:
* The fix works as intended.
* The only difference I noticed it that without the patch the player just shrugs whereas with the patch he knocks over the pillar (tested in chapter 1). Is there any condition in the vanilla game that the patch doesn't have anymore?
* When I type `test 50` in the Definitive Edition mod, nothing happens (the player won't be teleported to the pillar).

szapp:
Thanks for testing. Everything sounds great:

>     * The only difference I noticed it that without the patch the player just shrugs whereas with the patch he knocks over the pillar (tested in chapter 1). Is there any condition in the vanilla game that the patch doesn't have anymore?
The pillar is rotated by 180 degrees in the vanilla game. To knock it over, you'd need to stand on the other side (close to the cliff). It's very difficult to take the pillar in focus from that angle, however, and it takes a few times.

>     * When I type `test 50` in the Definitive Edition mod, nothing happens (the player won't be teleported to the pillar).
That's an indication, that the fix was not applied: The issue is already addressed. The reason for the teleport is not happening, is that the modified pillar is not found (because it is not modified).

AmProsius:
> That's an indication, that the fix was not applied: The issue is already addressed. The reason for the teleport is not happening, is that the modified pillar is not found (because it is not modified).

I understand. I only was confused, because for #87 the teleportation happened (even in the Definitive Edition) and for this fix it didn't. I would appreciate it, if the teleportation would happen regardless of the outcome. Or, as an alternative, a console output that the fix was skipped (for whatever reason).

szapp:
Okay, seems reasonable. I will quickly add this - although it might be unreliable in the case, if there are multiple columns in a mod. But since it's only for testing that's not an issue.

For some background: The reason that the teleport works in #87, is that it's tied to a way point in the world, whereas there are no way points close by to the pillar. So instead the hero is teleported to the pillar. The name of the pillar is adjusted while performing the fix (to indicate whether it has to be reverted later on). When the fix is not applied the teleport fails, as there is no object in the world with that name.

There is actually already an output when the test "fails" due to no teleportation, but only in the zSpy. I can fix the teleport and add a screen print that the fix was not applied.

AmProsius:
Thanks for the explanation. You don't need to fix the teleport for me, but a screen print would be nice. If that's done, I will merge this PR.

szapp:
> 
> 
> Thanks for the explanation. You don't need to fix the teleport for me, but a screen print would be nice. If that's done, I will merge this PR.

Done. Give it a try.

PR #98, opened 2021-02-01
#43: Fix whitespace of learning print (EN)

### LINKED ISSUE #43: Missing whitespace for skill point(s)
When learning from teachers, there is a whitespace missing before "skill point" and "skill points" respectively.

szapp:
### Description
Fixes #43. The screen information about learning skills now contains proper whitespace. This fix is for English localization only and should be tested for English (unfixed), English (fixed) and for another localization, where the fix does not apply.

### Test
Run automatic test with `test 43`. The test will throw an error if the localization is not English, but will pass.

### Additional Info
I only added the fix to the English section of the changelog.md.


AmProsius:
I've tested the fix. All tests pass:
1. The issue is still present in the unfixed game.
2. The issue is solved in the fixed game.
3. The fix is ignored in the German version of the game.

ScreenShot_2021_2_2_23_0_52

PR #100, opened 2021-02-02
#60: Fix Jesse's quest

### LINKED ISSUE #60: Jesse's quest not available
Jesse's quest PayForMe is not available due to a bug.

szapp:
### Description
Fixes #60. Jesse's quest can now be triggered.

### Test
Run automatic test with `test 60`.

### Additional Info
The quest does not have a name and is also not tracked in the quest log. Therefore its (non-existing) name is not listed in the changelog.


AmProsius:
The test passes, but when I played the quest line, I noticed more inconsistencies. Jesse asks the player to pay his 10 Ore to Bloodwyn. But even so the dialog line states that he does, the player doesn't lose 10 Ore.

It's no problem for me that the quest has no log entry as the whole Bloodwyn storyline isn't logged.

szapp:
Good catch. Should be an easy fix. I think it should be a separate issue, though. (Not conceptually, but implementation-wise). Could you create one?

PR #105, opened 2021-02-08
#8: Prevent NPCs to move in circles

### LINKED ISSUE #8: NPCs move in circles
NPCs are stuck or move in circles when trying to align if another NPC is standing on their spot.

szapp:
### Description
Fixes #8. NPCs should no longer move in circles when aligning in their daily routines.

### Test
Run manual test with `test 8`.

There is no working test for this issue yet, because I don't know where to reproduce this bug. We need to know a position in the world and time of day where the bug happens every time. See [here](https://github.com/AmProsius/gothic-1-community-patch/issues/8#issuecomment-774713219).

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix008_NpcMoveCircle.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix008_NpcMoveCircle.d
@@ -1,0 +1,104 @@
+/*
+ * #8 NPCs move in circles
+ */
+func int Ninja_G1CP_008_NpcMoveCircle() {
+    var int applied1; applied1 = FALSE;
+    var int applied2; applied2 = FALSE;
+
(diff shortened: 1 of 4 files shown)


szapp:
Just a note for future reference. The structure and names in the scripts changed considerably. Before merging this PR needs some adjustments, primarily the prefix in the function names.

szapp:
The actual bug described in #8 is that NPCs do not find their free-points and therefore walk in circles (if I understood Catalin correctly). This PR, however, is addressing something completely different: The continuous re-alignment of NPCs to their state free-points. (We could not even see the bug in the game).

I am not entirely sure, but it seems that issue and PR are addressing two different problems. 

@AmProsius, should we close this PR for now? We could revisit it if the re-alignment in fact needs fixing.

PR #106, opened 2021-02-09
#49: Fix the item text of the dungeon key

### LINKED ISSUE #49: Dungeon Key description faulty
"des Alten Lagers" from Dungeon Key should be translated to English.

szapp:
### Description
Fixes #49. Interestingly, I found a way to fix this issue completely language independent. The exact wording does not matter: if the second line of text overwrites the first line, it will be corrected.

### Test
Run automatic test with `test 49`.

The test, however, only supports English and German localization. Both should be tested (I only tested English). Three additional tests are yet to be performed.

1. Test in German localization (especially important because of the umlaut 'ö' in the key text)
1. Do an additional _manual_ test where the key was in the world before the fix was applied, i.e. create a game save with the key inserted, then apply the fix, load the game and manually confirm that the key's text was updated.
1. Do a test on a mod that already fixed the issue.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -24,6 +24,7 @@
 * Fix #38: Snaf now offers the dialog about Nek even when the quest "Snaf's Recipe" is completed.
 * Fix #40: Aleph doesn't offer to sell the key anymore when the player already obtained it.
 * Fix #43: The dialog choices about learning skills now contain proper whitespace.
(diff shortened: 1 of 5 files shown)


AmProsius:
All tests passed:
1. The key has the correct description in the German version of the game.
2. After loading a savegame with the key already inserted, the description was correct.
3. The description also was correct when tested with the UnPatch mod.

ScreenShot_2021_2_9_19_54_17

PR #107, opened 2021-02-09
#21: Prevent Fletcher from reopening quest

### LINKED ISSUE #21: Fletcher reopens closed quest
Fletcher reopens the quest "Vanished Guard" if the quest is already completed.

szapp:
### Description
Fixes #21. Fletch now leaves the quest closed if it is already.

### Test
Run automatic test with `test 21`.

**Not ready for merge:** The German translation of the changelog is missing.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -14,7 +14,7 @@
 * Fix #18: Bloodwyn doesn't ask for protection money anymore if the player has already joined a camp.
 * Fix #19: Scorpio can no longer be asked "Can you teach me to fight?" after he fled from the Old Camp to Cavalorn. Note: The player can still learn how to fight from him since that is a different dialog option.
 * Fix #20: Kirgo now correctly gives a beer to the player.
-* Fix #21: Fletcher now no longer reopens the quest "Vanished Guard".
+* Fix #21: Fletcher no longer reopens the quest "Vanished Guard" if it was closed before talking to him.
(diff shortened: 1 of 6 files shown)


szapp:
I actually want to change the implementation of the fix. Instead of checking the (possibly non existing) variable `Sly_LostNek`, I want to check the log status of the quest `CH1_LostNek` directly. This will more robust.

AmProsius:
> I actually want to change the implementation of the fix. Instead of checking the (possibly non existing) variable `Sly_LostNek`, I want to check the log status of the quest `CH1_LostNek` directly. This will be more robust.

I don't know if it makes a difference, but in Snaf's dialog te check is also against `Sly_LostNek`:
https://github.com/AmProsius/gothic-1-community-patch/blob/b8996cb83c3157b047d6109a3b203c357bc9bebb/scriptbase/_work/Data/Scripts/Content/Story/Missions/DIA_VLK_581_Snaf.d#L296-L309

szapp:
> I don't know if it makes a difference, but in Snaf's dialog te check is also against `Sly_LostNek`:

Yes, I every quest is accompanied by a variable as a proxy to check whether a quest is closed or opened. The status of the quests themselves cannot be checked (only set) from the scripts. I have made such a function now.

You are right. Checking the status of the quest itself does not really make a difference. The advantage here is that it is robust to changes of the variable and quest name.

PR #113, opened 2021-02-12
#78: Humans recognize orcs as enemies

### LINKED ISSUE #78: Humans don't recognise orcs as monsters
Humans ignore orcs until they're attacked by them. It's an old oversight from the alpha stage of the game where more interaction with orcs was planned.

szapp:
### Description
Fixes #78. Human NPCs now recognize orcs as enemies. This should fix the quest in the orc grave yard.

### Test
Run manual test with `test 78`.

Additionally to the test suite, the fix should be check with Baal Lukor in the orc grave yard. I have not done that yet.


AmProsius:
I've checked Baal Lukor and now he attacks correctly. He admittedly deals no damage, but he also doesn't receive damage as he's immortal.

PR #120, opened 2021-02-13
#11: Fix passage guard reaction for drawn weapon

### LINKED ISSUE #11: Player can pass guards in combat mode
The player can pass gate guards just by entering combat mode.

szapp:
### Description
Fixes #11.

The fix [suggested here](../issues/11#issuecomment-757133978) is not really tackling this specific problem of reacting to a drawn weapon, but rather:
1. It disables the reaction for **all** NPCs (not just the guards!) if the player was just warned by a passage guard.
1. It disables the reaction for **any** NPC that has important infos (a dialog)

Here the issues that arise from the above:
1. If the player draws a weapon near any guarded gate, *no* NPC will react. As soon as the player walks a bit away from the gate, NPC would suddenly react. This is a very strange behavior.
2. I first thought this is a very important thing to address: NPCs with important infos should talk to the player even if the weapon is drawn. But it's actually a very bad change. This would trigger dialogs in very inconvenient situations, e.g. in fights. Additionally, if the player draws their weapon and is trapped in a dialog, NPCs around will still wait for the player to remove their weapon. Once the dialog is over, the player is attacked and could not do anything about it.

The suggested fix is, in my opinion, rather problematic.

What I did here is a bit more targeted: The fix is only affecting guards where the player cannot pass - no other NPC.

### Test
Run manual test with `test 11`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,7 @@
 * Fix #7: NPCs no longer practice weapon training without a weapon.
 * Fix #9: NPCs no longer freeze while fleeing.
 * Fix #10: Followers adjust their walking speed according to the player. However, the player must be in sight of the follower if the player starts running after walking.
+* Fix #11: The player can no longer bypass NPCs that are guarding a passage by drawing the weapon.
 * Fix #12: The player no longer receives double XP when killing an NPC with a ranged weapon after already beaten the NPC in melee combat.
 * Fix #15: The player doesn't lose Strength as part of the quest "Horatio the Peasant" if he had more than 100 Strength.
 * Fix #16: Thorus can't be bribed if the player has already obtained the permit to pass the guards. Also the option to bribe Thorus disappears after the player has bribed him.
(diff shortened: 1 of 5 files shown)


szapp:
I will leave a note on how to enable the fix for any NPC with important infos for future reference only.
```diff
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix011_PassGuardsCombatMode.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix011_PassGuardsCombatMode.d
@@ -69,18 +69,6 @@ func int Ninja_G1CP_011_PassGuardsCombatMode() {
             */

             applied = TRUE;
+
+            // Optionally, remove the check for "Npc_IsInState(self, ZS_GuardPassage)" before
+            // This disables the reaction to drawing a weapon for any NPC that has an important info for the player
+            var int p25; p25 = MEM_ReadByte(pos-25);
+            var int p30; p30 = MEM_ReadByte(pos-30);
+            if (MEM_ReadByte(pos-20) == zPAR_TOK_CALLEXTERN)
+            && (MEM_ReadInt(pos-19) == MEM_GetFuncId(Npc_IsInState))
+            && ((p25 == zPAR_TOK_PUSHINT)  || (p25 == zPAR_TOK_PUSHVAR))
+            && ((p30 == zPAR_TOK_PUSHINST) || (p30 == zPAR_TOK_PUSHINT) || (p30 == zPAR_TOK_PUSHVAR)) {
+                MEMINT_OverrideFunc_Ptr = pos-30;
+                MEMINT_OFTokPar(zPAR_TOK_JUMP, pos-10 - currParserStackAddress);
+            };
         };
     end;

```

PR #153, opened 2021-02-27
#33: Fix Gorn's dialog about Shrike's quest

### LINKED ISSUE #33: Shrike's Hut quest
**Describe the bug**
* Gorn can be told about Shrike's defeat even if the player doesn't know about the hut.
* The quest "Shrike's Hut" can't be closed if the player defeats Shrike without knowing from Gorn about the hut.

**Expected behavior**
The quest "Shrike's Hut" can be closed no matter in which order the player gets to know about the quest and defeats Shrike.

**Additional context**
German translation of the quest is *Shrike's Hütte*.

szapp:
### Description
Fixes #33. The dialogs around Shrike's quest should be fixed.

### Test
Run automatic test with `test 33`.


AmProsius:
After testing the quest, I'd like to change the implementation. Instead of changing the condition for Gorn, I'd rather change the condition for Shrike, because the log entries make no sense otherwise.

---

https://github.com/AmProsius/gothic-1-community-patch/blob/e2fea999152ad59f67bc9d02f36befc68a334843/scriptbase/_work/Data/Scripts/Content/Story/Missions/DIA_ORG_842_Shrike.d#L73-L79

changed to

```d
FUNC int  DIA_Shrike_GetLost_Condition()
{
    if (Npc_KnowsInfo(hero,DIA_Gorn_Hut))
    && (self.aivar[AIV_WASDEFEATEDBYSC])
    {
        return 1;
    };
};
```

szapp:
How about this instead (in case the quest is available another way in a mod).

```D
FUNC int  DIA_Shrike_GetLost_Condition()
{
    if ((Gorn_ShrikesHut == LOG_RUNNING) || (Npc_KnowsInfo(hero, DIA_Gorn_Hut)))
    && (self.aivar[AIV_WASDEFEATEDBYSC])
    {
        return 1;
    };
};
```

szapp:
Do you have a suggestion for the changelog entry? The current one wouldn't fit so well anymore.

AmProsius:
> Do you have a suggestion for the changelog entry? The current one wouldn't fit so well anymore.

I don't agree. For me it still fits perfectly well. Why do you think it doesn't fit anymore?

> How about this instead (in case the quest is available another way in a mod).

Good catch! I totally agree.

szapp:
> I don't agree. 

You are right. Defeating ≠ get lost dialog. 

I think we can keep it.

PR #162, opened 2021-02-27
#124: Exchange castle gate guard ID

### LINKED ISSUE #124: Gate guard doesn't close castle gate in chapter 4
**Describe the bug**
Grd_280_Gardist has ID 230 instead of 280. Therefore he uses the routines of another NPC and the inner gate of the OC stays open in chapter IV.

**Expected behavior**
Gate guard of the castle now has the correct routine and stands at the gate.

szapp:
### Description
Fixes #124. The daily routine of the guard is fixed. Nevertheless, the original problem is not. See [here](../issues/124#issuecomment-787094949). For the time being I will leave this PR as a draft until we address the underlying issue. If we do, it should be solved in a separate fix. This fix here is complete.

### Test
Run manual test with `test 124`. The test will break the game and it will not be possible to continue playing. Save the game beforehand.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -44,7 +44,7 @@
 * Fix #109: The player now correctly loses 10 ore if he chooses to pay protection money to Bloodwyn later.
 * Fix #111: The player now correctly loses 10 ore if he chooses to pay protection money to Jackal.
 * Fix #112: The player now correctly loses 10 ore if he chooses to pay protection money to Jackal later.
-* Fix #124: The gate guard now correctly closes the castle gate of the Old Camp in chapter 4.
+* Fix #124: The guard of the castle gate in the Old Camp now has the correct daily routine to watch over and operate the winch of the gate.
 
 ## DE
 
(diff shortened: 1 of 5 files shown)


szapp:
The issue about the blocked gate is addressed here #163. This PR is in itself now ready for merging.

docs/changelog.md
```diff
@@ -44,6 +44,7 @@
 * Fix #109: The player now correctly loses 10 ore if he chooses to pay protection money to Bloodwyn later.
 * Fix #111: The player now correctly loses 10 ore if he chooses to pay protection money to Jackal.
 * Fix #112: The player now correctly loses 10 ore if he chooses to pay protection money to Jackal later.
+* Fix #124: The gate guard now correctly closes the castle gate of the Old Camp in chapter 4.
```
szapp, line 47: This description depends on #163 now. Should we rephrase it?

AmProsius: *Gate guard of the castle now has the correct routine and stands at the gate.*

szapp: Thanks, I added a bit more detail, see 3b8d863.

PR #164, opened 2021-02-27
#163: Unblock the castle gate

### LINKED ISSUE #163: Castle gate of the Old Camp cannot close
**Describe the bug**
The castle gate of the Old Camp is blocked and cannot close.

**Expected behavior**
The castle gate of the Old Camp is no longer blocked and can properly close.

**Additional context**
The issue arose from a discussion in #124.

1. **Positioning**
The gate is blocked through the wall by the winch (the switch/wheel) that is used to open it. Unfortunately, because of the size and orientation of the bounding boxes the winch has to be moved quite far to avoid collision between the gate and the winch. There are two options:
    1. Either the winch is moved to the opposite wall (see screenshot below)
    2. Or the winch stays where it is and we simply turn off its dynamic collision. That would have the side-effect that the player can walk through it (using it is still possible).

2. **Gate state stored in game save**
Aside from that there is a more pressing issue: The state of the gate is stored in the game save. Once we unblock the gate and it is closed, we will have to revert it to be open on saving the game. The problem here is, that we don't know if the gate was opened during the game due to our fixes (see #124) or by the mod. Therefore, we cannot just revert the gate to *open* on saving, because that might break the mod. Are there any suggestions how we could tackle that problem?

winch_reposition

szapp:
### Description
Fixes #163. Unblock the castle gate of the Old Camp to let it properly close.

### Test
Run manual test with `test 163`.

### Additional Information
* Currently the winch is moved to the opposite wall to remove its collision with the gate. There is an alternative mentioned in #163. 
* If the gate is closed, its state is written to the game save. So far it is not handled yet to revert the gate's state before saving.

The points above should be discussed/adjusted before merging this PR. Therefore it is marked as draft for the time being.


szapp:
The fix is now converted to a session fix and is ready for review/testing/merging.

PR #167, opened 2021-02-28
#46: Fix the key instance of the smithy's door

### LINKED ISSUE #46: Door near the smithy in the castle of the Old Camp is inaccessible
**Describe the bug**
The smith's door cannot not be opened, because it references a key that does not exist.

**Expected behavior**
The door near the smithy in the castle of the Old Camp is now accessible with the correct key which can now be obtained in Stone's cell in the dungeon of the Old Camp.

**Additional context**
The key to the smithy's door does not work because the door references a wrong key instance name: `ITKEY_OB_SMITH_01` does not exist. It needs to be changed to `ItKe_OB_Smith_01`.

door

**Implementation**
This issue was merged with #47. We supply our own G1CP copy of the key instance to allow saving and loading. When removing G1CP the then invalid instance will automatically vanish on loading. If reinstalling G1CP the key instance will be re-added (given the conditions to do so are met).

### Apply
```
1. Is door valid (identified,
   locked, has incorrect key instance)
     │             │
     │ True        │ False
     │             └─────────────────────────────────────────────┐
     V                                                           │
2. No such key instance exists                                   │
   (could have been added by a mod to                            │
    fix the issue the other way around)                          │
     │             │                                             │
     │ True        │ False                                       │
     │             └──────────────────────────────────────────┐  │
     V                                                        │  │
3. Does the G1CP key already exist in the world               │  │
     │             │                                          │  │
     │ True        │ False                                    │  │
     │             │                                          │  │
     │             V                                          │  │
     │        A. Can key be inserted?                         │  │
     │           (waypoint exists)                            │  │
     │             │            │                             │  │
     │             │ True       │ False                       │  │
     │             │            └──────────────────────────┐  │  │
     │             V                                       │  │  │
     │        B. Is story progressed enough?               │  │  │
     │           (ExploreSunkenTower == True)              │  │  │
     │             │            │                          │  │  │
     │             │ True       │ False                    │  │  │
     │             │            │                          │  │  │
     │             │            V                          │  │  │
     │             │     I.  Will story variable be set    │  │  │
     │             │         in expected function?         │  │  │
     │             │         (ExploreSunkenTower           │  │  │
     │             │          mentioned in                 │  │  │
     │             │          B_Story_ExploreSunkenTower)  │  │  │
     │             │            │           │              │  │  │
     │             │            │ True      │ False        │  │  │
     │             │            │           └───────────┐  │  │  │
     │             │            V                       │  │  │  │
     │             │     II. Hook after the story       │  │  │  │
     │             │         function to run this here  │  │  │  │
     │             │         function again and to      │  │  │  │
     │             │         insert the key then        │  │  │  │
     │             │            │                       │  │  │  │
     │             │            └────────────────────┐  │  │  │  │
     │             │                                 │  │  │  │  │
     │             │                                 │  │  │  │  │
     │             V                                 │  │  │  │  │
     │        C. Insert the key                      │  │  │  │  │
     │           (in Stone's cell)                   │  │  │  │  │
     │             │           │                     │  │  │  │  │
     │             │ Success   │ Failed (unlikely)   │  │  │  │  │
     │             │           └──────────────────┐  │  │  │  │  │
     │             │                              │  │  │  │  │  │
     V             V                              │  │  │  │  │  │
4. Adjust the key instance                        │  │  │  │  │  │
   for the door                                   │  │  │  │  │  │
        │                                         │  │  │  │  │  │
        │                                         │  │  │  │  │  │
        V                                         V  V  V  V  V  V
    return True                                      return False
    (fix applied)                                 (fix not applied)
```

### Revert
```
1. Is door valid (identified, has
   updated G1CP key instance)
     │
     │ True
     V
2. Adjust the key instance
   for the door

(G1CP key vanishes automatically if loading a game save without G1CP active)
```

In short, the fix (i.e. adjusting the door) will **only** be applied if
1. the door has the incorrect key instance
2. there is no such key instance in the script
3. the story has progressed far enough

Would the story be changed somewhat (the variables renamed) the entire fix would not be applied.

szapp:
### Description
Fixes #46. Make the smithy's door accessible.

### Test
Run manual test with `test 46`. The test has two passes.
1. First function call: The PC is teleported to Stone's cell and the story is advanced to chapter 4.
2. Second function call: The PC is teleported in front of the door.

There are some tests to check off.
- [x] The key shows up in Stone's cell (first pass)
- [x] The key opens the door (second pass)
- [x] The key is still present in the inventory after saving, quitting the game, restarting and loading
- [x] The key is still present in the world (drop it on the ground) after saving, quitting the game, restarting and loading
- [x] The key is still present in a chest after saving, quitting the game, restarting and loading
- [x] **The key is gone when removing G1CP and loading (there should be no visible error)** *most interesting one*
- [x] The key is properly re-inserted when re-installing G1CP (note that this only works for game saves where the door was not unlocked yet).

To understand when the key is and is not inserted, the [flow-chart](../issues/46#issue-780608458) may be consulted (or even verified).


szapp:
I think there is still a problem. When loading a game save twice the door might not be locked again. I will have a look before merging.

szapp:
I removed re-locking of the door. The reasoning:

> When reverting, the door will not be locked again. The reasoning behind this is, that G1CP is only supplying a way of unlocking the door. It's the player that may unlock it. G1CP will only revert its own actions, not the consequences.

I hope we can agree on this.

AmProsius:
Do I understand correctly that the key won't fit anymore, but it remains open, because the player unlocked it before (if he did so)? If so, I don't see any problem as the saves won't be affected.

PR #149, opened 2021-03-20
Spelling - Improved Ore Armor (DE/EN)

AmProsius:
**Describe the bug**
In the German and English localization there's a typo in the Improved Ore Armor's name.

**Changelog**
* Den Namen "verbesserte Erzrüstung" korrigiert zu "Verbesserte Erzrüstung".
* Changed the name "Improved ore Armor" to "Improved Ore Armor".

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
@@ -1,0 +1,16 @@
+/*
+ * #149 Spelling - Improved Ore Armor (DE/EN)
+ */
+func int G1CP_149_DE_EN_ImprovedOreArmorName() {
+    var int symbId; symbId = MEM_GetSymbolIndex("ORE_ARMOR_H");
+    const string needle = "";
+    const string replace = "";
+    if (G1CP_Lang == G1CP_Lang_DE) {
+        needle  = "verbesserte Erzrüstung";
+        replace = "Verbesserte Erzrüstung";
+    } else if (G1CP_Lang == G1CP_Lang_EN) {
+        needle  = "Improved ore Armor";
+        replace = "Improved Ore Armor";
+    };
+    return (G1CP_ReplaceAssignStr(symbId, "C_Item.name", 0, needle, replace) > 0);
+};
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
@@ -3,8 +3,8 @@
  */
 func int G1CP_149_DE_EN_ImprovedOreArmorName() {
     var int symbId; symbId = MEM_GetSymbolIndex("ORE_ARMOR_H");
-    const string needle = "";
-    const string replace = "";
+    const string needle = "G1CP invalid string";
+    const string replace = "G1CP invalid string";
     if (G1CP_Lang == G1CP_Lang_DE) {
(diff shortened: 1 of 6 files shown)


src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
```diff
@@ -0,0 +1,16 @@
...
+    const string needle = "";
+    const string replace = "";
+    if (G1CP_Lang == G1CP_Lang_DE) {
+        needle  = "verbesserte Erzr�stung";
+        replace = "Verbesserte Erzr�stung";
+    } else if (G1CP_Lang == G1CP_Lang_EN) {
+        needle  = "Improved ore Armor";
+        replace = "Improved Ore Armor";
+    };
+    return (G1CP_ReplaceAssignStr(symbId, "C_Item.name", 0, needle, replace) > 0);
```
szapp, line 15: Something that I now realize is that we should be careful here. If the language is neither English nor German, the instance function will be searched for `name = "";` and replaced with `name = "";`. Of course, this is not a problem and also not very likely to happen, but we should keep that in mind for future uses of `G1CP_ReplaceAssignStr` that the needle and replace string should be non-empty if possible.

src/Ninja/G1CP/Content/Tests/test149.d
```diff
@@ -0,0 +1,38 @@
+/*
+ * #149 Spelling - Improved Ore Armor (DE/EN)
+ *
+ * The text of the item "ORE_ARMOR_H" is inspected programmatically.
+ *
+ * Expected behavior: The armor will have the correct description (checked for English and German localization only).
```
szapp, line 6: Here it says "description" but it is the name in this fix. Maybe we should make these comments more generic. I think it is very tedious to adjust these comments every time (or potentially have to go back and fix them if we forgot). We can keep that in mind going forward.

src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
szapp: Thinking about it, an alternative implementation could have been
```D
func int G1CP_149_DE_EN_ImprovedOreArmorName() {
    var int symbId; symbId = MEM_GetSymbolIndex("ORE_ARMOR_H");
    const string needleEN  = "Improved ore Armor";
    const string replaceEN = "Improved Ore Armor";
    const string needleDE  = "verbesserte Erzrüstung"
    const string replaceDE = "Verbesserte Erzrüstung";
    var int count; count = 0;
    count += G1CP_ReplaceAssignStr(symbId, "C_Item.name", 0, needleEN , replaceEN);
    count += G1CP_ReplaceAssignStr(symbId, "C_Item.name", 0, needleDE , replaceDE);
    return (count > 0);
};
```
However, this is computationally less efficient.

For now, this fix is great as it is.

szapp (approved):

src/Ninja/G1CP/Content/Fixes/Session/fix149_DE_EN_ImprovedOreArmorName.d
szapp: We could instantiate the strings non-empty with highly unlikely content:

```D
    const string needle = "G1CP invalid string";
    const string replace = "G1CP invalid string";
```

szapp (changes requested):
The test fails for German localization. It seems the file test149.d is not encoded in Windows-1252. And the string "Verbesserte Erzrüstung" (including the umlaut) won't match. Otherwise all good.

szapp (approved):
Test passes for German and English localization

PR #146, opened 2021-03-23
Spelling - Novice's Loincloth (DE)

AmProsius:
**Describe the bug**
In the German localization there's a typo in the Novice's Loincloth's name.

**Changelog**
Den Namen "Novizen Rock" korrigiert zu "Novizenrock".

### DIFF
--- a/docs/changelog_de.md
+++ b/docs/changelog_de.md
@@ -3,6 +3,7 @@
 ## v1.1.0 (TBA)
 ### General
 * Fix #144: Die Rüstung "Gomez'Rüstung" heißt nun korrekt "Gomez' Rüstung".
+* Fix [#146](https://github.com/AmProsius/gothic-1-community-patch/pull/146): Die Rüstung "Novizen Rock" heißt nun korrekt "Novizenrock".
 * Fix #147: Die Rüstung "Crawler-Plattenrüstung" heißt nun korrekt "Crawlerplatten-Rüstung".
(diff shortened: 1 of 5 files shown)


docs/changelog_de.md
```diff
@@ -3,6 +3,7 @@
 ## v1.1.0 (TBA)
 ### General
 * Fix #144: Die Rüstung "Gomez'Rüstung" heißt nun korrekt "Gomez' Rüstung".
+* Fix [#146](https://github.com/AmProsius/gothic-1-community-patch/pull/146): Die Rüstung "Novizen Rock" heißt nun korrekt "Novizenrock".
```
szapp, line 6: I was against links in the changelog at first, because it would become hard to read if displayed in plain text. But I now think that is very unlikely and people will mostly view this file fully formatted here on Github.

Linking to the details of each fix is a really nice idea, to lower the threshold for users to get into the issue tracker on Github and for them to get more details in general. I would propose, however, to shorten the link to https://g1cp.org/issues/146. That link works irrespective of whether the fix is listed as issue or has been turned into a pull request.

Should we go ahead and update the entire changelog?

szapp (approved):
I have a minor suggestion regarding the link in the changelog. I have not run the test yet. Once I did I will merge the PR.

docs/changelog_de.md
AmProsius: > Should we go ahead and update the entire changelog?

I'd like to do so. I already wrote that on my todo list.

szapp (approved):

PR #142, opened 2021-03-28
Dialog lines with ambient templar interchanged pt. 1

szapp:
**Describe the bug**
In the German localization a dialog line with an ambient templar is interchanged.

**Changelog**
Vertauschte Dialogzeilen mit einem Ambient-Templer korrigiert.

**Additional context**
See #197 for pt. 2.


szapp:
Note, that I did not prefix the function names with "DE_" because this fix does not involve any language strings. It therefore also works irrespective of localization, e.g. if the English scripts of a mod were ported from the German ones including this bug.

AmProsius (commented):
Fix approved.

PR #52, opened 2021-03-28
Exploit: Grindstone in the New Camp can be used without sword blade

szapp:
**Describe the bug**
The grindstone in the New Camp can be used without having a sword blade in the inventory.

**Expected behavior**
The grindstone in the New Camp now correctly requires a sword blade to use.

**Additional context**
It's currently possible to produce *Grobe Schwerter* without having Swordblades in the inventory.  
This issue was split into two further issues #212 #213.  
Add `useWithItem` with `ItMiSwordblade` to grindstone at `NC_HUT03_OUT_MOVEMENT`.

**Screenhots**
ScreenShot_2021_2_4_23_1_41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.1.0 (TBA)
 ### General
+* Fix [#52](https://g1cp.org/issues/52): The grindstone in the New Camp now correctly requires a sword blade to use.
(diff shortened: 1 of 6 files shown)


szapp:
The test is manual. The player is teleported to the grindstone and should no longer be able to use it without a sword blade. What I have not tested yet, is that this change is reverted properly.

AmProsius (commented):
The fix is reverted correctly when removing the patch.

PR #51, opened 2021-03-29
Chest in stonehenge crypt behind wall 

szapp:
**Description**
There is a chest in the crypt that's behind a wall.

**Expected behavior**
Moved chest in the crypt from outside to inside the focus room to make it available to the player.

**Additional context**

4E76550D-F30D-4567-AE50-BCA1B6CA41F4

2AD15B58-4A88-4C24-9741-CA4E33DADB5E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.1.0 (TBA)
 ### General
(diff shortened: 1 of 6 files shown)


szapp:
The test requires manual confirmation. I already checked if the chest will be re-positioned on save and load.

PR #181, opened 2021-03-29
Swiney gives his own pants to the player

szapp:
**Describe the bug**
When Swiney gives a Digger's Dress to the player, he gives him his own equipped Digger's Dress.

**Expected behavior**
Swiney no longer loses his own Digger's Dress when he gives one to the player.


szapp:
I fixed the issue differently from discussed here: Instead of changing his "giving behavior", he merely re-equips his best armor right after the **execution** of the dialog. This should be instant and not visible.

I could not see him take off his armor. @AmProsius, could you test if you spot him taking it off and putting it on? The test is manual (but the dialog will be immediately available).

PR #193, opened 2021-03-29
Gates triggered by switches are stuck after loading

szapp:
**Describe the bug**
After using a switch/lever/winch to close a gate, the switch does not work anymore after loading a game and the gate is stuck.

**Expected behavior**
Objects activated by switches/levers/winches are no longer stuck after loading a game.

1. Close a gate by using the respective switch.
2. Save the game.
3. Load the game.
4. The switch state returns to 0 while the gate remains in its triggered state: The gate will be stuck for good.

**Additional context**
[Attached](https://www.sendspace.com/file/49pe45) the fix done by szapp 2 years ago.  
See also the original post on WoG: https://forum.worldofplayers.de/forum/threads/?p=25954985


szapp:
The test requires manual confirmation:
> The gate (that the PC is teleported to) can be opened again after closing it, saving and loading.

Note: The state of the fix (i.e. `APPLIED` or `NOT APPLIED`) reflects whether any stuck switches where fixed. If none, the fix will be listed as `NOT APPLIED`.

PR #23, opened 2021-03-29
Baal Namib sells Novice's Armor too late

szapp:
**Describe the bug**
Baal Namib doesn't sell the Novice's Armor until chapter 3, but the player can already become a Templar in chapter 2.

**Expected behavior**
Baal Namib now sells the Novice's Armor in chapter 2 already, if the Minecrawler Queen quest is running or completed.


szapp:
This is an automatic test for a change. The only restrictions for obtain the armor is now:
- Does not have the armor (unchanged)
- Guild is novice (unchanged)
- Enough ore (unchanged)
- Chapter >= 2 (**new**)

PR #218, opened 2021-03-30
Improve fisk fence quest

### LINKED ISSUE #36: Fisk's quest isn't triggered when Mordrag is dead
**Describe the bug**
Fisk's quest "New Fence for Fisk" isn't triggered when Mordrag is dead (either killed by the player or by a monster).

**Expected behavior**
Fisk's quest "New Fence for Fisk" is now also available if Mordrag is dead.

szapp:
This PR constraints the conditions introduced in v1.1.0 in #36. Now the fence quest is only available if Mordrag was knocked down (default) or if he is dead. The additional option of him being in the New Camp was removed here.

Additionally, the fix was reimplemented to fix a major bug of overwriting the global variable `MordragKO_HauAb`, which would cause inwanted side-effects in other dialogs/story.

### Description
Fixes #36.

### Test
Run automatic test with `test 36`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix [#192](https://g1cp.org/issues/192): Mages (NPCs fighting only with spells) no longer auto-equip weapons (e.g. after trading). This requires fix #59 to be active.
 
 ### Story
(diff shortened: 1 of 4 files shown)


AmProsius (approved):
Everything works as expected.

PR #14, opened 2021-03-30
The player can cancel fights by entering a room 

szapp:
**Describe the bug**
The player can run into a room and immediately leave it to cancel a fight if the NPC has full HP.

**Expected behavior**
The player can no longer cancel fights by entering and leaving a room.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.1.0 (TBA)
 ### General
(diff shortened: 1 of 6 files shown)


szapp:
The test is manual:
- Cause a fight with an NPC (without causing them damage!) and enter and leave the nearby room: The NPC will continue the fight.
- Enter a room and wait for an NPC to react and attack the PC. Leave the room (without causing them damage!). The NPC will stop the fight.

AmProsius:
> @AmProsius can you give me a scenario how to trigger it reliably?

I reproduced the bug by attacking the immortal gate guard at the main gate of the Old Camp and then running into and out of a hut. The guard then stopped attacking me and returned to the gate.

PR #182, opened 2021-03-31
Exploit: The gate guard can be bribed multiple times with Dusty as a follower

szapp:
**Describe the bug**
When bringing Dusty to the Swamp Camp, the player can indefinitely bribe the gate guard to get EXP.

**Expected behavior**
The gate guard can only be bribed once when bringing Dusty to the Swamp Camp.

1. Bribe the gate guard with Dusty following you.
2. Save the game
3. Load the save.
4. Bribe the gate guard again (if you have enough ore).


szapp:
The test is automatic. It may still be worth to check it manually (with saving and loading as written in the issue description). I have not tested that yet.

Issue #211, opened 2021-03-28
Bugs found in the original game using zSpy (DE/EN/PL/EN)


N1kX94:
These points are in the world.zen is not present.
Warn:  0 U:   SPAWN: Spawnpoint OW_MONSTER_NAVIGATE02 not found. Npc SNAPPER cannot be spawned. .... <oSpawn.cpp,#386>
Warn:  0 U:   SPAWN: Spawnpoint OW_MONSTER_NAVIGATE02 not found. Npc SNAPPER cannot be spawned. .... <oSpawn.cpp,#386>

There is this point OW_MONSTER_NAVIGATE_02
Need replace in STARTUP_SUB_SURFACE
OW_MONSTER_NAVIGATE02  -> OW_MONSTER_NAVIGATE_02

N1kX94:
In one chest there is an extra one ,
Warn:  0 B:     MOBCONTAINER: Item-Instance not found: , .... <oMobInter.cpp,#2627>
		[% oCMobContainer:oCMobInter:oCMOB:zCVob 64513 16024]
			pack=int:0
			presetName=string:
			bbox3DWS=rawFloat:-35454.0898 2280.677 -15369.6855 -35365.7734 2357.57202 -15269.3906 
			trafoOSToWSRot=raw:ceb5a63e00000000a40c723f000000000000803f00000000a40c72bf00000000ceb5a63e
			trafoOSToWSPos=vec3:-35412.4219 2280.677 -15320.7129
			vobName=string:CHEST
			visual=string:CHESTBIG_OCCHESTMEDIUMLOCKED.MDS
			showVisual=bool:1
			visualCamAlign=enum:0
			cdStatic=bool:0
			cdDyn=bool:1
			staticVob=bool:1
			dynShadow=enum:0
			[visual zCModel 0 16025]
			[]
			[ai % 0 0]
			[]
			focusName=string:CHEST
			hitpoints=int:10
			damage=int:0
			moveable=bool:0
			takeable=bool:0
			focusOverride=bool:0
			soundMaterial=enum:0
			visualDestroyed=string:
			owner=string:
			ownerGuild=string:
			isDestroyed=bool:0
			stateNum=int:1
			triggerTarget=string:
			useWithItem=string:
			conditionFunc=string:
			onStateFunc=string:
			rewind=bool:0
			locked=bool:-1
			keyInstance=string:ItKe_Focus5
			pickLockStr=string:
[...]

PR #214, opened 2021-03-31
Graham doesn't sit at campfire

szapp:
**Describe the bug**
Graham is supposed to sit at the campfire in the evening, but a typo in the waypoint name prevents him from doing so.

**Expected behavior**
Graham now correctly sits at the campfire in the evening.

**Additional context**
Bug and fix provided by [NikX94](https://github.com/AmProsius/gothic-1-community-patch/issues/211#issuecomment-808914450).


szapp:
The test requires manual confirmation. It may happen that Graham still stands around, if the other digger happens to chose to sit at Graham's spot. That's not so likely (only if the PC stands in the way during their change in routines, at 6 PM).

PR #13, opened 2021-03-31
NPCs draw weapon on dead monsters 

szapp:
NPCs sometimes draw their weapon even if the monster is already dead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v1.1.0 (TBA)
 ### General
+* Fix [#13](https://g1cp.org/issues/13): NPCs no longer draw their weapons on dead monsters.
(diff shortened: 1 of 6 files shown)


szapp:
The test is not functional yet. I cannot reproduce the problem, and need an example of how to trigger the bug.

szapp:
This bug still has to be validated. Whoever does this, when doing so, please provide instruction on how to reproduce the bug, such that the test can be adjusted.

szapp:
Moved to v1.2.0, because of unclear circumstances of the bug.

szapp:
What's left to do here is to clarify the circumstances that cause this bug. I could not reproduce it. As I never experienced it I don't know when it happens. Someone else will have to provide that information. I will have to remove myself from being assigned.

PR #110, opened 2021-03-31
Weird camera distance when sitting on a bench

szapp:
**Describe the bug**
When the player is sitting on a bench, the camera zooms too much into the player.

**Expected behavior**
When the player is sitting on a bench, the camera no longer zooms too much into the player.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v1.1.0 (TBA)
 ### General
 * Fix [#52](https://g1cp.org/issues/52): The grindstone in the New Camp now correctly requires a sword blade to use.
+* Fix [#110](https://g1cp.org/issues/110): Mouse input is no longer locked when interacting with objects to allow adjusting the camera viewpoint.
 * Fix [#149](https://g1cp.org/issues/149): The armor "Improved ore Armor" is now correctly labelled as "Improved Ore Armor".
 * Fix [#192](https://g1cp.org/issues/192): Mages (NPCs fighting only with spells) no longer auto-equip weapons (e.g. after trading). This requires fix #59 to be active.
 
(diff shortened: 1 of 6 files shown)


szapp:
I added a preliminary implementation to try it out and get a feel for it. If we choose to go with this option (unlocking the mouse input), the fix should probably be re-implemented with machine code. Until we have decided how to fix the issue, this here is very much work in progress and should not be merged!

AmProsius:
> I would like to discuss whether to just replace the camera distance values for the bench (a direct bug fix, but given they are unchanged in Gothic 2, not the correct bug fix) or to unlock mouse input (going beyond a bug fix and with the risk of unintended side effects for other MOB camera definitions).

Given that Gothic 2 changed the mouse behavior, I would like to go that path. The new values seem too opinionated for me. If we unintentionally create a new bug with enabling the mouse wheel, we can always go back to the fixed values.

szapp:
> > I would like to discuss whether to just replace the camera distance values for the bench (a direct bug fix, but given they are unchanged in Gothic 2, not the correct bug fix) or to unlock mouse input (going beyond a bug fix and with the risk of unintended side effects for other MOB camera definitions).
> 
> Given that Gothic 2 changed the mouse behavior, I would like to go that path. The new values seem too opinionated for me. If we unintentionally create a new bug with enabling the mouse wheel, we can always go back to the fixed values.

Okay, sounds reasonable. I was working on the alternative (with explicit values). To not throw that away, I pushed that on a separate branch: [bug110-alternative](../tree/bug110-alternative).

Now I will try to re-implement the solution here in machine code to make it feasible.

szapp:
I managed to implement it in a very light way. This fix should be tested thoroughly! That is, with many different MOBs to see if the newly allowed camera rotations do not interfere with anything. The implementation is on machine code level and now matches that of Gothic 2 quite closely.

To get it working, the mouse setting for MOBs is now hijacking the setting of ladders. I increased the X-movement from 0.5x speed to 1x speed (compared to the normal camera movement), because the ladder-camera is quite slow to move.

AmProsius (commented):
I couldn't find any problems with MOBs. It's really cool to be able to move the camera now!

PR #132, opened 2021-04-01
Spelling - Book: "Die Gruft" (DE)

AmProsius:
**Describe the bug**
In the German localization there's a wrong description for the book "The Crypt".

**Changelog**
Die Beschreibung des Buches "Die Gruft" von "Kampfkunst" korrigiert zu "Die Gruft".

### DIFF
--- a/docs/changelog_de.md
+++ b/docs/changelog_de.md
@@ -5,6 +5,7 @@
 * Fix [#4](https://g1cp.org/issues/4): Zuvor besiegte NSCs fahren den Spieler nun nicht mehr an, wenn er ihnen im Weg steht.
 * Fix [#50](https://g1cp.org/issues/50): Die unerreichbare Truhe in der Gruft unter dem Stonehenge ist nun korrekt positioniert und zugänglich.
 * Fix [#52](https://g1cp.org/issues/52): Der Schleifstein im Neuen Lager setzt zur Benutzung nun korrekt eine Schwertklinge voraus.
+* Fix [#132](https://g1cp.org/issues/132): Das Buch "Die Gruft" heißt nun korrekt "Die Gruft" statt "Kampfkunst".
 * Fix [#144](https://g1cp.org/issues/144): Die Rüstung "Gomez'Rüstung" heißt nun korrekt "Gomez' Rüstung".
 * Fix [#145](https://g1cp.org/issues/145): Die Rüstung "leichte Söldnerrüstung" heißt nun korrekt "Leichte Söldnerrüstung".
 * Fix [#146](https://g1cp.org/issues/146): Die Rüstung "Novizen Rock" heißt nun korrekt "Novizenrock".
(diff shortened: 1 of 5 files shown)


szapp:
Is this book actually in the game at all? It is the only document in the game that you cannot open (= read).

I think we cannot really sell something as a "fix" that doesn't have any impact on the game at all. @AmProsius, any thoughts on this?

docs/changelog_de.md
```diff
@@ -5,6 +5,7 @@
 * Fix [#4](https://g1cp.org/issues/4): Zuvor besiegte NSCs fahren den Spieler nun nicht mehr an, wenn er ihnen im Weg steht.
 * Fix [#50](https://g1cp.org/issues/50): Die unerreichbare Truhe in der Gruft unter dem Stonehenge ist nun korrekt positioniert und zugänglich.
 * Fix [#52](https://g1cp.org/issues/52): Der Schleifstein im Neuen Lager setzt zur Benutzung nun korrekt eine Schwertklinge voraus.
+* Fix [#132](https://g1cp.org/issues/132): Die Beschreibung des Buchs "Die Gruft" lautet nun korrekt "Die Gruft" statt "Kampfkunst".
```
szapp, line 8: I updated the changelog here to distinguish the description from the name of the item.

szapp (commented):
Is this book actually in the game at all? It is the only document in the game that you cannot open (= read).

I think we cannot really sell something as a "fix" that doesn't have any impact on the game at all. @AmProsius, any thoughts on this?

szapp:
Let me correct myself: The item is placed in the right most room upstairs of the ore baron's house on the table next to the candle stick and the book `Kampfkunst`. However, it does not appear in the game. I cannot see it or pick it up.

szapp:
I checked all worlds and found two other item-VOBs without instance. In all cases the item either does not exist in the scripts or its placement is questionable. The scroll does not exist in the scripts, the cello-item is commented out in the scripts (i.e. does not exist), and the book has no content as discussed above. I think we should ignore all of them, including this book here.

| World         | Item Instance       | Nearest Waypoint                    |
| :------------ | :------------------ | :---------------------------------- |
| WORLD.ZEN     | DIE_GRUFT           | OCC_BARONS_UPSTAIRS_RIGHT_BACK_EXIT |
|               | ITMICELLO           | SPAWN_OW_BLACKGOBBO_A1              |
|               |                     |                                     |
| ORCTEMPEL.ZEN | ITARSCROLLMASSDEATH | TPL_276                             |

AmProsius:
The issue is related to https://g1cp.org/issues/211#issuecomment-80902682. Either we restore the book or we remove any trails.

szapp:
Ah I hadn’t seen that this issue was already mentioned there. I think we should put this fix here (the spelling issue) on hold until decided if we want to make the item available in the game. Because I would like to avoid having too much unnecessary code being parsed when loading the patch. 

As for the invalid item, I could imagine that the item was invalidated on purpose last minute, because the book has no content. Instead of removing the VOB from the world (which would have required to edit the world in the spacer), its instance/visual was just quickly removed. But that is just a wild guess and it may just have been a mistake. 

What I find troublesome is that the book has no content, whereas all documents in the game (to my knowledge) can be opened and viewed. Adding this empty book item has therefore no worth or contribution, but will also add a bit of inconsistency. That’s my impression, but I am open for discussion.

AmProsius:
> I think if we decide to not restore the book, we don’t need to do anything, because it is not noticeable in the game.

Blubbler [states](https://forum.worldofplayers.de/forum/threads/1574630-Release-Gothic-1-Community-Patch/page6?p=26726598&viewfull=1#post26726598) that the invisible items an still be picked up in the right angle. I have to validate that behavior, but if that's true, I'd like to remove the invisible book from the game. This PR can be closed regardless of the outcome of that test.

PR #115, opened 2021-04-01
Dummy Templars survive in orc graveyard

szapp:
**Describe the bug**
When the player is fast enough, he can kill the orcs near the dummy Templars before the orcs kill the Templars. Then the dummy Templars just stand there doing nothing.

**Expected behavior**
The dummy Templars in the orc graveyard are now already dead when the player enters the graveyard.

**Additional context**
My suggestion would be to kill the dummy Templars when the player first enters the orc graveyard.

**Screenshot**
ScreenShot_2021_2_12_21_37_49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix [#23](https://g1cp.org/issues/23): The armor "Novice's Armor" is now available in chapter 2 to make it accessible before receiving the first templar armor.
 * Fix [#36](https://g1cp.org/issues/36) (updated): Fisk's quest "New Fence for Fisk" is now available when the player either knocked out or killed Mordrag.
 * Fix [#37](https://g1cp.org/issues/37): Gravo is now correctly listed as merchant in the Old Camp in the journal.
+* Fix [#115](https://g1cp.org/issues/115): The templars in the orc graveyard are now less likely to survive.
 * Fix [#174](https://g1cp.org/issues/174): The key "Gomez' Bowl" is now correctly labelled as "Gomez' Key".
 * Fix [#175](https://g1cp.org/issues/175): The key "Rice Lord's Bowl" is now correctly labelled as "Rice Lord's Key".
 * Fix [#181](https://g1cp.org/issues/181): Swiney no longer takes off his own Digger's Dress when he gives one to the player.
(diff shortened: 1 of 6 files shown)


szapp:
I committed the changes so far, if you want to try it out. Unfortunately, I didn't write a test yet, but entering the orc graveyard (irrespective of the story progression suffices).

AmProsius:
Perfect, go ahead then.

AmProsius:
Fun fact: This is the painting "Sitting Duck" by Bedard that might have influenced the state:

3A80E87C-00B6-4DC0-9DDA-573ABE18F9D6

I have a copy of that painting in my living room.

AmProsius (commented):
Really cool fix! I enjoyed running through the orc graveyard and seeing the orcs kill the templars. It's way more immersive than my proposed solution.

szapp:
Some notes that apparently were filtered by Github:

> What about "bleeding" as in health degeneration + blood pfx effect?

and

> >"That "sitting duck" AI-state is some comments in the code describing, that is exclusively used for NPCs to be killed of (I quoted that comment a few posts above). Therefore, the changes are only beneficial in all cases."
> A mod author could have used the state without having read the comments. Killing all "sitting duck" characters in all mods semms risky to me.

*(Copied from [here](https://forum.worldofplayers.de/forum/threads/?p=26726598))*

It seems unlikely to me for anyone to use the AI state at all, even less so for a different purpose. The sitting ducks really react to nothing at all. They have no perceptions and wouldn’t even react to the player talking to them. That they are waiting to be killed is even in the name of the AI state. That’s my justification to killing off any NPC in that state if the player comes to close.  
Nonetheless, I am all for being rather on the safe side instead of making assumptions. So, thanks for the feedback! I guess we could add checks to only alter the AI state for the NPC we specify (the handful templars) as initially planned. If a mod modifies/exchanges or adds more of them, that’s on the mod developer. 


Regarding the blood effect, I don’t understand the question/comment. Could you elaborate, please?

AmProsius:
> Regarding the blood effect, I don’t understand the question/comment. Could you elaborate, please?

I think he meant something like the effect when the player character gets hit by an arrow/bolt. I wouldn't include that though, I'm happy with the current solution.

The concerns regarding the use of the "Sitting Duck" AI state in mods are reasonable, so I'd like to add checks for the inserted templars. On the other hand, this fix wouldn't apply to additional templars that could be added by a mod then. I'm still in favor of the additional checks though.

PR #185, opened 2021-04-08
Scorpio offers to teach crossbow multiple times

AmProsius:
**Describe the bug**
Scorpio offers to teach crossbow level 1 multiple times, even if the player already gained that skill.

**Expected behavior**
Scorpio no longer offers to teach crossbow level 1 if the player already gained that skill.

1. Learn crossbow level 1 (from him).
2. Save the game.
3. Load the save.
4. Scorpio offers to teach crossbow level 1 again.


szapp (approved):
I updated the fix and test with the new symbol functions. All good, test passes!

PR #186, opened 2021-04-08
Corristo can be asked to join the mages multiple times

AmProsius:
**Describe the bug**
The player can ask Corristo to become a mage multiple times, even if he already met the level requirement before.

**Expected behavior**
Corristo's dialog option to become a mage now disappears when the player has already asked him before and met the level requirements.

1. Ask Corristo to become a mage.
2. Save the game.
3. Load the save.
4. Ask Corristo to become a mage again.


AmProsius:

src/Ninja/G1CP/Content/Fixes/Session/fix186_CorristoJoinMages.d
```diff
@@ -0,0 +1,27 @@
...
+};
+
+/*
+ * This function intercepts the dialog condition to introduce more conditions
+ */
+func int G1CP_186_CorristoJoinMages_Hook() {
+    G1CP_ReportFuncToSpy();
+
+    // Add the new condition (other conditions remain untouched)
+    if (G1CP_GetIntVar("Corristo_KDFAufnahme", 0, FALSE) == TRUE) {
```
szapp, line 21: I will remove the check against `TRUE` here, as a mod may set this variable to another non-zero value (for whatever reason).

szapp (approved):
I updated getting the symbols with the new functions. The test passes.

PR #188, opened 2021-04-08
Cor Kalom can be told that the weed was delivered multiple times

### LINKED ISSUE #24: Cor Kalom closes wrong quest
Cor Kalom closes the "Weed monopoly" quest instead of "Weed delivery".

AmProsius:
**Describe the bug**
The player can tell Cor Kalom that he delivered the weed multiple times.

**Expected behavior**
Cor Kalom can now be told that the weed was delivered only once.

1. Tell Cor Kalom that the weed was delivered.
2. Save the game.
3. Load the save.
4. Tell Cor Kalom about the weed again.

**Additional context**
Fix for #24 has to be applied to successfully apply this fix.


src/Ninja/G1CP/Content/Fixes/Session/fix188_CorKalomWeedExploit.d
```diff
@@ -0,0 +1,44 @@
...
+
+    // Check if quest is running
+    var int symbPtr; 
+    if (G1CP_IsFixApplied(24)) {
+        symbPtr = MEM_GetSymbol("Kalom_Krautbote");
+    } else {
+        symbPtr = MEM_GetSymbol("Kalom_DrugMonopol");
+    };
+    if (symbPtr) {
+        if (MEM_ReadInt(symbPtr + zCParSymbol_content_offset) == LOG_SUCCESS) {
```
szapp, line 37: I suggest taking the inverse of
```D
if (var == LOG_RUNNING) {
    return TRUE;
};
```
So instead I propose to turn this condition here into

```D
if (var != LOG_RUNNING) {
    return TRUE;
};
```

szapp (approved):
I did not explicitly trace back the relation to #24, but the test passes if #24 is applied and also if it's not.

AmProsius:
> I did not explicitly trace back the relation to #24, but the test passes if #24 is applied and also if it's not.

Yes, I managed to not require the fix for #24 to be applied.

For example:
```d
if (G1CP_IsFixApplied(24)) {
    symbId = G1CP_GetIntVarID("Kalom_Krautbote", 0);
} else {
    symbId = G1CP_GetIntVarID("Kalom_DrugMonopol", 0);
};
```

PR #230, opened 2021-04-11
Add getters for testsuite

AmProsius:
Add first set of getters for tests.

### DIFF
--- a/src/Ninja/G1CP/Content/Tests/test003.d
+++ b/src/Ninja/G1CP/Content/Tests/test003.d
@@ -13,7 +13,7 @@ func void G1CP_Test_003() {
     var int arrowId; arrowId = G1CP_Testsuite_CheckItem("ItAmArrow");
     G1CP_Testsuite_CheckItem("Diegos_Bogen");
     G1CP_Testsuite_CheckItem("ItRw_Bow_Small_01");
-    G1CP_Testsuite_CheckFunc("B_RegainDroppedWeapon", "void|C_Npc", "");
+    G1CP_Testsuite_CheckFunc("B_RegainDroppedWeapon", "void|inst", "");
     G1CP_Testsuite_CheckPassed();
 
     // Insert test NPC
--- a/src/Ninja/G1CP/Content/Testsuite/testsuite.d
+++ b/src/Ninja/G1CP/Content/Testsuite/testsuite.d
@@ -27,6 +27,7 @@ const int G1CP_TestsuiteStatusPassed = TRUE;
  */
 func int G1CP_Testsuite() {
     CC_Register(G1CP_TestsuiteList, "test list", "List all tests of G1CP");
+    CC_Register(G1CP_TestsuiteNext, "test next", "Execute next manual test of G1CP");
     CC_Register(G1CP_TestsuiteAll, "test all", "Run complete test suite for G1CP");
     CC_Register(G1CP_TestsuiteCmd, "test ", "Run test from test suite for G1CP");
 
@@ -176,6 +177,51 @@ func string G1CP_TestsuiteCmd(var string command) {
     G1CP_TestsuitePrintErrors();
 
     return retStr;
+};
+
+func string G1CP_TestsuiteNext(var string _) {
+    // Constant to continue where left off
+    const int symbId = 0;
+
+    // Restart from the beginning
+    if (symbId <= 0) || (symbId >= G1CP_SymbEnd) {
+        symbId = G1CP_SymbStart;
+    };
+
+    // Iterate over symbols and find next manual test
+    while(symbId < G1CP_SymbEnd);
+        // Compare symbol name
+        var zCPar_Symbol symb; symb = _^(MEM_GetSymbolByIndex(symbId));
+        if (STR_StartsWith(symb.name, "G1CP_TEST_"))
+        && (STR_Len(symb.name) == 13)
+        && ((symb.bitfield & zCPar_Symbol_bitfield_type) == zPAR_TYPE_FUNC) {
+            // Get test ID
+            var int id; id = STR_ToInt(STR_SubStr(symb.name, 10, 3));
+
+            // Check if manual or automatic
+            if (symb.offset == (zPAR_TYPE_VOID >> 12)) {
+                break;
+            };
+        };
+        symbId += 1;
+    end;
+
+    // Reached last
+    if (symbId >= G1CP_SymbEnd) {
+        return "Reached end of manual tests. Will restart from the beginning on next call";
+    };
+
+    // Otherwise execute test and advance iterator
+    G1CP_TestsuiteCmd(IntToString(id));
+    symbId += 1;
+
+    // Return info
+    var string msg; msg = "Start test of #";
+    msg = ConcatStrings(msg, IntToString(id));
+    msg = ConcatStrings(msg, " '");
+    msg = ConcatStrings(msg, G1CP_GetFixShortName(id));
+    msg = ConcatStrings(msg, "'");
+    return msg;
 };
 
 func string G1CP_TestsuiteList(var string _) {
--- a/src/Ninja/G1CP/Content/Tests/test004.d
+++ b/src/Ninja/G1CP/Content/Tests/test004.d
@@ -17,5 +17,6 @@ func void G1CP_Test_004() {
     G1CP_NpcSetAIVarI(npc, aiVarId, TRUE);
 
     // Exactly in the way
+    Wld_SetTime(12,0);
     AI_Teleport(hero, wp.name);
 };
--- a/src/Ninja/G1CP/Content/Tests/test037.d
+++ b/src/Ninja/G1CP/Content/Tests/test037.d
@@ -7,13 +7,10 @@
  */
 func int G1CP_Test_037() {
     const string TEMP_TOPIC_NAME = "G1CP Test 37"; // Has to be a unique name with absolute certainty
-    var int topicId; topicId = G1CP_Testsuite_CheckStringConst("GE_TraderOC", 0);
+    const string GE_TraderOC = ""; GE_TraderOC = G1CP_Testsuite_GetStringConst("GE_TraderOC", 0);
     var int infoId; infoId = G1CP_Testsuite_CheckInfo("DIA_Gravo_HelpHow");
     var int funcId; funcId = G1CP_Testsuite_CheckDialogFunc("DIA_Gravo_HelpHow_Info");
     G1CP_Testsuite_CheckPassed();
-
-    // Get constant values
-    const string TOPIC = ""; TOPIC = G1CP_GetStringConstI(topicId, 0, TOPIC);
 
     // Check status of the test
     var int passed; passed = TRUE;
@@ -21,19 +18,19 @@ func int G1CP_Test_037() {
     // First test: Check if the dialog function creates the topic if it did not exist beforehand
 
     // Rename the log topic if it already exists
-    G1CP_LogRenameTopic(TOPIC, TEMP_TOPIC_NAME);
+    G1CP_LogRenameTopic(GE_TraderOC, TEMP_TOPIC_NAME);
 
     // Just run the dialog and see what happens
     G1CP_Testsuite_Call(funcId, 0, 0, TRUE);
 
     // Check if the log entry was created
-    if (!G1CP_LogGetTopic(TOPIC)) {
+    if (!G1CP_LogGetTopic(GE_TraderOC)) {
         G1CP_TestsuiteErrorDetail("Log topic was not created by the dialog function");
         passed = FALSE;
     };
 
     // Clean up
-    G1CP_LogRemoveTopic(TOPIC);
+    G1CP_LogRemoveTopic(GE_TraderOC);
 
     // Second test: Check if the log entry is created on applying the fix
     // This second test is specific to how the G1CP fixes the bug and will fail if a mod fixed the issue in another way
@@ -52,7 +49,7 @@ func int G1CP_Test_037() {
     r = G1CP_037_LogEntryGravoMerchant();
 
     // Check if the topic was created
-    if (!G1CP_LogGetTopic(TOPIC)) {
+    if (!G1CP_LogGetTopic(GE_TraderOC)) {
         G1CP_TestsuiteErrorDetail("Log topic was not auto-created on applying the fix");
         passed = FALSE;
     };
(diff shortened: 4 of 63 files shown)


src/Ninja/G1CP/Content/Testsuite/getter.d
```diff
@@ -0,0 +1,120 @@
...
+ */
+func int G1CP_Testsuite_GetInfoId(var string name) {
+    // Check if info instance exists
+    if (!G1CP_IsInfoInst(name)) {
+        var string msg; msg = ConcatStrings("Dialog function '", name);
+        msg = ConcatStrings(msg, "' not found");
+        G1CP_TestsuiteErrorDetail(msg);
+        G1CP_TestsuiteStatusPassed = FALSE;
+    };
+
```
szapp, line 88: Not 100% sure, but it might be worth to add an additional check for `G1CP_GetInfo(name) != 0` in here. This determines if the info instance is listed in the infomanager (i.e. the dialog is available in the game).

src/Ninja/G1CP/Content/Testsuite/getter.d
```diff
@@ -0,0 +1,120 @@
...
+};
+
+/*
+ * Check if integer variable exists
+ */
+func void G1CP_Testsuite_CheckIntVar(var string name, var int arrIdx) {
+    // Check if variable exists
+    if (!G1CP_IsIntVar(name, arrIdx)) {
+        var string msg; msg = ConcatStrings("Integer variable '", name);
+        msg = ConcatStrings(msg, "' not found/valid");
```
szapp, line 41: Maybe we can go back to "not found" and instead of  "not found/valid". It should suffice that the kind of symbol was checked in the beginning of the message, e.g. "Integer variable".

src/Ninja/G1CP/Content/Testsuite/getter.d
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    // Return the item's symbol index
+    return MEM_GetSymbolIndex(name);
+};
+
+/*
+ * Check if integer variable exists and return its symbol pointer
+ */
+func int G1CP_Testsuite_GetIntVarPtr(var string name, var int arrIdx) {
```
szapp, line 20: I think we don't need that function. `G1CP_Testsuite_CheckIntVar` will probably suffice. The pointer is not required, because we can read/write the variable using `G1CP_GetVar` and `G1CP_SetVar`.

src/Ninja/G1CP/Content/Testsuite/getter.d
```diff
@@ -0,0 +1,120 @@
...
+        msg = ConcatStrings(msg, "' not found/valid");
+        G1CP_TestsuiteErrorDetail(msg);
+        G1CP_TestsuiteStatusPassed = FALSE;
+    };
+};
+
+/*
+ * Check if integer constant exists and return its symbol pointer
+ */
+func int G1CP_Testsuite_GetIntConstPtr(var string name, var int arrIdx) {
```
szapp, line 50: See above (G1CP_Testsuite_GetIntVarPtr).

src/Ninja/G1CP/Content/Testsuite/testsuite.d
```diff
@@ -275,3 +284,18 @@ func string G1CP_TestsuiteList(var string _) {
 
     return ret;
 };
+
+func void G1CP_Testsuite_CheckPassed() {
```
szapp, line 288: We should add a header comment here with a brief description of the function (see other functions of this file).

src/Ninja/G1CP/Content_G1.src
```diff
@@ -32,7 +32,8 @@ Content\Misc\npc.d
 
 // Fixes and tests
 Content\Fixes\record.d
-Content\testsuite.d
+Content\Testsuite\testsuite.d
+Content\Testsuite\getter.d
```
szapp, line 36: Since not all of the functions are getter functions now, should change the name of the file to something like `checks.d` or `conditions.d`?

szapp (commented):
Some thoughts

src/Ninja/G1CP/Content/Testsuite/getter.d
AmProsius: Whx not `if (!G1CP_GetInfo(name))`?

src/Ninja/G1CP/Content/Testsuite/checks.d
```diff
@@ -0,0 +1,119 @@
...
+
+    // Return the dialog's symbol index
+    return MEM_GetSymbolIndex(name);
+};
+
+/*
+ * Check if function exists
+ */
+func void G1CP_Testsuite_CheckFunc(var string name) {
+    if (!G1CP_IsFunc(name, "int|none")) {
```
szapp, line 84: The second argument here is the expected signature of the function (see comments in [func.d](../../blob/master/src/Ninja/G1CP/Content/Symbols/func.d)). As it is right now, it will only allow functions with an integer return type and no parameters.

I think we can have one general function with `func void G1CP_Testsuite_CheckFunc(string name, string signature)` and then special functions for reoccurring fixes: E.g. `G1CP_Testsuite_CheckDialogFunc` with ("void|none") and `G1CP_Testsuite_CheckDialogConditionFunc` with ("int|none") and maybe some more.

src/Ninja/G1CP/Content/Tests/test015.d
```diff
@@ -9,22 +9,17 @@ func int G1CP_Test_015() {
...
-    var int passed; passed = TRUE;
-
-    // Check if the dialog function exists
-    var int funcId; funcId = MEM_GetSymbolIndex("DIA_Horatio_HelpSTR_LEARN_NOW");
-    if (funcId == -1) {
-        G1CP_TestsuiteErrorDetail("Dialog function 'DIA_Horatio_HelpSTR_LEARN_NOW' not found");
-        return FALSE;
-    };
+    // Prior checks
+    var int funcId; funcId = G1CP_Testsuite_GetDialogFuncId("DIA_Horatio_HelpSTR_LEARN_NOW");
```
szapp, line 13: I think we won't need the `funcId` here actually. We only need to confirm the function's existence. Below we can use `MEM_CallByString("DIA_Horatio_HelpSTR_LEARN_NOW")` instead of `MEM_CallById(funcId)`. So `G1CP_Testsuite_CheckDialogFunc` would suffice. I think that will aid in cleaning up the tests and make them more readable.

src/Ninja/G1CP/Content/Tests/test015.d
```diff
@@ -60,5 +53,5 @@ func int G1CP_Test_015() {
     hero.attribute[ATR_STRENGTH] = strengthBak;
 
     // At last
-    return passed;
+    return result;
```
szapp, line 56: I like the variable name `passed` here better actually. (I think I had suggested the name "result", but "passed" is more descriptive.)

src/Ninja/G1CP/Content/Testsuite/checks.d
```diff
@@ -100,13 +100,35 @@ func void G1CP_Testsuite_CheckDialogFunc(var string name) {
     G1CP_Testsuite_CheckFunc(name, "void|none", "Dialog function");
 };
 
+/*
+ * Check if dialog function exists and return its symbol index
+ */
+func int G1CP_Testsuite_GetDialogFuncId(var string name) {
```
szapp, line 106: We might not need this function, see above.

src/Ninja/G1CP/Content/Testsuite/checks.d
```diff
@@ -100,13 +100,35 @@ func void G1CP_Testsuite_CheckDialogFunc(var string name) {
...
  * Check if dialog condition function exists
  */
 func void G1CP_Testsuite_CheckDialogConditionFunc(var string name) {
     G1CP_Testsuite_CheckFunc(name, "int|none", "Dialog condition function");
 };
 
+/*
+ * Check if dialog condition function exists and return its symbol index
+ */
+func int G1CP_Testsuite_GetDialogConditionFuncId(var string name) {
```
szapp, line 124: We might not need this function, see above.

src/Ninja/G1CP/Content/Tests/test015.d
szapp: That may go for all function getter-functions.

src/Ninja/G1CP/Content/Tests/test007.d
```diff
@@ -70,23 +65,17 @@ func int  ZS_G1CP_Test_007_NpcRountine_Loop() {
...
-        if (symbId != -1) {
-            // AI_StartState(self, symbId, 0, ""); // Does not work, expects func parameter
-            MEM_PushInstParam(self);
-            MEM_PushIntParam(symbId); // Func parameter as integer
-            MEM_PushIntParam(0);
-            MEM_PushStringParam("");
-            MEM_Call(AI_StartState);
+        // AI_StartState(self, symbId, 0, ""); // Does not work, expects func parameter
+        MEM_PushInstParam(self);
+        MEM_PushIntParam(MEM_GetSymbolIndex("ZS_PracticeSword")); // Func parameter as integer
```
AmProsius, line 70: Wouldn't it be better to use `MEM_PushIntParam(funcId);` and save the return value of `G1CP_Testsuite_CheckAiStateStartFunc("ZS_PracticeSword");` in line 14?

szapp: Would have been nice, but it's not within the same function.

src/Ninja/G1CP/Content/Tests/test003.d
```diff
@@ -6,9 +6,15 @@
...
  */
 func void G1CP_Test_003() {
-    if (!G1CP_TestsuiteAllowManual) {
-        return;
-    };
+    G1CP_Testsuite_CheckManual();
+    G1CP_Testsuite_CheckAiStateStartFunc("ZS_PracticeSword");
+    var int symbId1; symbId1 = G1CP_Testsuite_CheckItem("Whistlers_Schwert");
+    var int symbId2; symbId2 = G1CP_Testsuite_CheckItem("Wolfs_Bogen");
+    var int symbId3; symbId3 = G1CP_Testsuite_CheckItem("ItAmArrow");
```
AmProsius, line 13: To add readability, I would rename the variables to `swordId`, `bowId` and `arrowId` respectively.

szapp: But don't we loose copy&paste-bility with this again?

AmProsius: With the numbers behind `symbId` you'd have to rename the variable anyway.

src/Ninja/G1CP/Content/Tests/test025.d
```diff
@@ -6,17 +6,13 @@
...
-    // Define constants for specific test
-    const string GUILD_NAME = "GIL_KDW";
-
-    // Prior checks
-    var int funcId; funcId = G1CP_Testsuite_GetDialogConditionFuncId("KDW_600_Saturas_HEAVYARMOR_Condition");
-    var int itemId; itemId = G1CP_Testsuite_GetItemId("KDW_ARMOR_H");
-    G1CP_Testsuite_CheckIntConst(GUILD_NAME, 0);
+    var int funcId; funcId = G1CP_Testsuite_CheckDialogConditionFunc("KDW_600_Saturas_HEAVYARMOR_Condition");
+    var int itemId; itemId = G1CP_Testsuite_CheckItem("KDW_ARMOR_H");
+    var int gilId; gilId = G1CP_Testsuite_CheckIntConst("GIL_KDW", 0);
```
AmProsius, line 11: I'd suggest `guildId`, because `gilId` is too German.

src/Ninja/G1CP/Content/Tests/test026.d
```diff
@@ -6,33 +6,28 @@
  * Expected behavior: The condition function will return FALSE.
  */
 func int G1CP_Test_026() {
+    var int funcId; funcId = G1CP_Testsuite_CheckDialogConditionFunc("Info_Org_804_FirstWarn_Condition");
+    var C_Npc npc; npc = G1CP_Testsuite_FindNpc("Org_804_Organisator");
+    var int gilId; gilId = G1CP_Testsuite_CheckIntConst("GIL_GRD", 0);
```
AmProsius, line 11: I'd suggest `guildId`, because `gilId` is too German.

src/Ninja/G1CP/Content/Tests/test183.d
```diff
@@ -6,17 +6,13 @@
...
-    // Define constants for specific test
-    const string GUILD_NAME = "GIL_KDF";
-
-    // Prior checks
-    var int funcId; funcId = G1CP_Testsuite_GetDialogConditionFuncId("KDF_402_Corristo_HEAVYARMOR_Condition");
-    var int robeId; robeId = G1CP_Testsuite_GetItemId("KDF_ARMOR_H");
-    G1CP_Testsuite_CheckIntConst(GUILD_NAME, 0);
+    var int funcId; funcId = G1CP_Testsuite_CheckDialogConditionFunc("KDF_402_Corristo_HEAVYARMOR_Condition");
+    var int robeId; robeId = G1CP_Testsuite_CheckItem("KDF_ARMOR_H");
+    var int gilId; gilId = G1CP_Testsuite_CheckIntConst("GIL_KDF", 0);
```
AmProsius, line 11: I'd suggest `guildId`, because `gilId` is too German.

src/Ninja/G1CP/Content/Testsuite/find.d
```diff
@@ -0,0 +1,50 @@
...
+        } else {
+            _^(infoPtr);
+        };
+    } else {
+        MEM_NullToInst();
+    };
+};
+
+/*
+ * Check if waypoint exists and return its
```
AmProsius, line 39: Typo in comment: `... return it`.

src/Ninja/G1CP/Content/Testsuite/get.d
```diff
@@ -0,0 +1,35 @@
...
+};
+
+/*
+ * Check if string variable exists and return it
+ */
+func int G1CP_Testsuite_GetStringVar(var string name, var int arrIdx) {
+    var int varId; varId = G1CP_Testsuite_CheckStringVar(name, arrIdx);
+
+    return G1CP_GetStringVarI(varId, arrIdx, 0);
+};
```
szapp, line 26: This function should return a string and the fallback value (third argument) of `G1CP_GetStringVarI` should be a string as well. Same for `G1CP_Testsuite_GetStringConst` below.

szapp:
All tests seem to still pass.

PR #37, opened 2021-03-26
Gravo not listed as merchant 

szapp:
Gravo isn't listed in "Diary/General info/Merchant in the Old Camp" if the topic wasn't previously created.

PR #204, opened 2021-04-16
Cavalorn not listed as tutor

AmProsius:
**Describe the bug**
When asking Fingers about Cavalorn, he is not listed in "Diary/General Info/Tutor outside the camps" if the topic wasn't previously created.

**Expected behavior**
When asking Fingers about Cavalorn, he is now listed in "Tutor outside the camps" in the journal.

**Additional context**
Similar to #37 and #203.


AmProsius:

AmProsius:
Currently I get `Log topic was not auto-created on applying the fix` when running `test 204`, but I don't have a clue why. Also the fix is not applied.

szapp:
This fix is special, because the dialog is permanent. It's told status will always be false. That means two things:
1. We cannot add the dialog entry post-hoc, since we don't know whether the dialog has been told in the game before.
2. The log entry should be infinite, if I am not mistaken, because the dialog can be triggered indefinitely.

szapp:
So this fix will have to be implemented differently:
1. Intercept `B_LogEntry` and create the log topic if it did not exist before (as done in the current implementation already).
2. But also check if that entry already exists and prevent adding it again.
3. Nothing will be applied reverted on saving and loading, because it cannot be checked whether the dialog has been spoken or not. This fix will thereby be no longer a revertible fix.

I will make the necessary adjustments.

szapp:
I rewrote the fix. Please not the commit message body above, as to why I adjusted the changelog as well.

szapp:
@AmProsius I cannot assign you as reviewer as you opened the pull request.

PR #203, opened 2021-04-16
Graham not listed as merchant

AmProsius:
**Describe the bug**
Graham is not listed in "Diary/General Info/Merchant in the Old Camp" if the topic wasn't previously created.

**Expected behavior**
Graham is now listed in "Merchant in the Old Camp" in the journal.

**Additional context**
Similar to #37.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Gamesave/fix203_LogEntryGrahamMerchant.d
+++ b/src/Ninja/G1CP/Content/Fixes/Gamesave/fix203_LogEntryGrahamMerchant.d
@@ -1,0 +1,149 @@
+/*
+ * #203 Graham not listed as merchant
+ */
+
+/*
+ * Since applying and reverting both depend on finding the necessary symbols this is a toggle function
+ */
+func int G1CP_203_LogEntryGrahamMerchant_Toggle(var int apply) {
+    // Define possibly missing symbols locally
+    const int LOG_MISSION = 0;
+    const int LOG_NOTE    = 1;
+
+    // Parameters of this fix
+    const int    topicSection  = LOG_NOTE;
+    const string topicSymbName = "GE_TraderOC";
+    const string infoSymbName  = "DIA_Graham_Hello";
+    const string funcSymbName  = "DIA_Graham_Hello_Info";
+    const string hookSymbName  = "G1CP_203_LogEntryGrahamMerchant_Intercept";
+
+    // Retrieve the topic and entry strings once and modify the info function
+    const int    infoId  = -2; // -1 is reserved for invalid symbols
+    const int    topicId = -1;
+    const int    entryId = -1;
+    const string topic   = "G1CP invalid topic name";
+    const string entry   = "G1CP invalid topic entry";
+    if (infoId == -2) {
+        infoId = MEM_GetSymbolIndex(infoSymbName);
+        topicId = MEM_GetSymbolIndex(topicSymbName);
+        var int funcId; funcId = MEM_GetSymbolIndex(funcSymbName);
+        var int b_logentry_id; b_logentry_id = MEM_GetSymbolIndex("B_LogEntry");
+
+        // Do this only once (this is never reverted, i.e. session fix)
+        if (infoId == -1) || (funcId == -1) || (b_logentry_id == -1) || (topicId == -1) {
+            return FALSE;
+        };
+
+        // Find all calls to 'B_LogEntry' within the dialog function
+        const int bytes[2] = {zPAR_TOK_CALL<<24, -1};
+        var zCPar_Symbol needleSymb; needleSymb = _^(MEM_GetSymbolByIndex(b_logentry_id));
+        bytes[1] = needleSymb.content;
+        var int matches; matches = G1CP_FindInFunc(funcId, _@(bytes)+3, 5);
+
+        /* We are looking for:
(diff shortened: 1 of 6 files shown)


AmProsius:

src/Ninja/G1CP/Content/Fixes/Gamesave/fix203_LogEntryGrahamMerchant.d
```diff
@@ -0,0 +1,149 @@
...
+    const int    infoId  = -2; // -1 is reserved for invalid symbols
+    const int    topicId = -1;
+    const int    entryId = -1;
+    const string topic   = "G1CP invalid topic name";
+    const string entry   = "G1CP invalid topic entry";
+    if (infoId == -2) {
+        infoId = MEM_GetSymbolIndex(infoSymbName);
+        topicId = MEM_GetSymbolIndex(topicSymbName);
+        var int funcId; funcId = MEM_GetSymbolIndex(funcSymbName);
+        var int b_logentry_id; b_logentry_id = MEM_GetSymbolIndex("B_LogEntry");
```
szapp, line 30: These lines are something we can refactor soon. Let's merge this PR first, and then do the entire scripts at once. To use more specific functions instead of `MEM_GetSymbolIndex`.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix203_LogEntryGrahamMerchant.d
```diff
@@ -0,0 +1,149 @@
...
+        // Do this only once (this is never reverted, i.e. session fix)
+        if (infoId == -1) || (funcId == -1) || (b_logentry_id == -1) || (topicId == -1) {
+            return FALSE;
+        };
+
+        // Find all calls to 'B_LogEntry' within the dialog function
+        const int bytes[2] = {zPAR_TOK_CALL<<24, -1};
+        var zCPar_Symbol needleSymb; needleSymb = _^(MEM_GetSymbolByIndex(b_logentry_id));
+        bytes[1] = needleSymb.content;
+        var int matches; matches = G1CP_FindInFunc(funcId, _@(bytes)+3, 5);
```
szapp, line 41: I suppose I can create another convenience function. We have one for replacing function calls, but not one to just *find* function calls. That might come in handy. It would be something like `G1CP_FindCall`, analogous to the existing `G1CP_ReplaceCall`.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix203_LogEntryGrahamMerchant.d
```diff
@@ -0,0 +1,149 @@
...
+                entryId = MEM_ReadInt(pos-4);
+                break;
+            };
+        end;
+        MEM_ArrayFree(matches);
+
+        // Check if we have found a valid entry
+        if (entryId <= 0) || (entryId >= MEM_Parser.symtab_table_numInArray) {
+            return FALSE;
+        };
```
szapp, line 63: This if-condition can also be updated later. I will be much easier to use `G1CP_IsStringI(entryId)`.

src/Ninja/G1CP/Content/Tests/test203.d
```diff
@@ -0,0 +1,107 @@
...
+        G1CP_TestsuiteErrorDetail("Log topic was not auto-created on applying the fix");
+        passed = FALSE;
+    };
+
+    // Revert everything
+    r = G1CP_203_LogEntryGrahamMerchantRevert();
+    G1CP_SetInfoTold(infoSymbName, toldBak);
+    r = G1CP_203_LogEntryGrahamMerchant();
+    G1CP_LogRemoveTopic(topic);
+    G1CP_LogRenameTopic(tempTopicName, topic);
```
szapp, line 103: One thing to keep in mind with this test (in case we use it somewhere else in the future), is that it breaks the fix-lookup-table. Applying and reverting the fix like done here, does not update the fix status. In this particular function, the fix status will remain "applied" even the fix is not. The fix-lookup-table will be corrupted from that point on.
Maybe we could backup the current status of the fix in the beginning with `G1CP_GetFixStatus` and set it in the end with `G1CP_SetFixStatus`.

szapp (approved):
Very nice. I added some comments for later. But it's ready to merge (test passes).

src/Ninja/G1CP/Content/Fixes/Gamesave/fix203_LogEntryGrahamMerchant.d
szapp: Done with b25e153.

src/Ninja/G1CP/Content/Tests/test203.d
szapp: I think, since this is just a test, this is not worth it. I would suggest to keep in mind that tests may just break the game.

szapp: Since this is just a test, I think this is not so important, but feel free to fix it
```D
var int fixStatusBak; fixStatusBak = G1CP_GetFixStatus(203);

// ...

G1CP_SetFixStatus(203, fixStatusBak);
```
If you do fix this, there is some more tests where this can be done. But it's only applicable to tests where the fix-status is changed **inside the apply function** - this seems to only happen for these sorts of log entry fixes.

PR #205, opened 2021-04-16
Wolf not listed as merchant

AmProsius:
**Describe the bug**
Wolf is not listed in "Diary/General Info/Merchant in the New Camp" if the topic wasn't previously created.

**Expected behavior**
Wolf is now listed in "Merchant in the New Camp" in the journal.

**Additional context**
Similar to #37, #203 and #204.


AmProsius:

szapp (approved):

PR #45, opened 2021-04-16
Snappers don't spawn outside the Monastery ruins 

szapp:
Snappers don't spawn outside the Monastery ruins.

This fix corrects the spawn of the two snappers such that they appear in the game. On saving, this fix is reverted to leave no changes in the savegame. That means, the spawn of the snappers is invalidated again. However, if the player is currently in spawn range of the snappers when saving, their spawn is not reverted and they will be written to the savegame. In that event, they remain in the game. This is necessary to avoid inconsistencies when saving the game while engaging with the snappers.


szapp:
The test will have to be a bit elaborate by checking if the snappers are gone when unloading the patch.

PR #212, opened 2021-04-17
Cauldron in the New Camp can be used without a scoop pt. 1

szapp:
**Describe the bug**
The cauldron near the ore heap in the New Camp can be used without having a Scoop in the inventory.

**Expected behavior**
The cauldron near the ore heap in the New Camp now correctly requires a scoop to use.

**Additional context**
- This issue was split from #52. Part one of #213.
- Closest waypoint: `NC_HUT13_OUT`
- The `useWithItem` of the MOB should be `ITMISCOOP`
- The `sceme` of the MOB is `CAULDRON`
- Suggested short name: `UseWithItemNCCauldron1`

**Screenhots**
ScreenShot_2021_2_4_23_3_1


AmProsius (changes requested):
The player character now gets a scoop if he didn't have one before and uses the cauldron.

PR #176, opened 2021-04-18
The text of items with no value is incomplete in the inventory

szapp:
**Describe the bug**
There are items that have no value. In the inventory it shows "Value:" but not number.

**Expected behavior**
Items with no value no longer show an empty "Value:" in the inventory text.

**Additional context**
The items have a value of 0 and therefor the value isn't displayed.

The bug has been reported in #177 #178 #179 #180.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## v1.2.0 (TBA)
 ### General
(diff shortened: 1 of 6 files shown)


szapp:
The fix is implemented to remove the display of "Value:" (which is determined from the string constant `NAME_Value`) from the inventory display if the item value is zero. As of now, the implementation is sub-optimal, because it is a script hook. That is not good in this case, because the hook will be executed 5 times per frame when the inventory is open. I will rewrite it in machine code.

PR #194, opened 2021-04-24
NPCs do not collect the weapon after defeating an opponent

szapp:
**Describe the bug**
NPCs do not reliably pick up the weapon of their defeated opponent.

**Expected behavior**
NPCs now correctly pick up the weapon and ammunition of their defeated opponent.

1. Attack an NPC.
2. Get defeated by the NPC in a fight.
3. The NPC might loot the ore of the PC but not will fail to pick up the weapon.


szapp:
The test is semi-automatic. It is manual, but the success will be displayed on the screen. Failure will be visible from the zSpy.

PR #58, opened 2021-04-25
Player can decelerate falls by fight actions in mid-air

szapp:
When the player falls down (e.g. a cliff) and has a weapon drawn, he can decelerate the fall by continuously blocking in mid-air.


szapp:
The fix (and test) are complete, but I noticed, that *any* fight action can interrupt falling. Therefore, it may makes sense to change the fix. Note: check 0x613F3D in `oCAIHuman::PC_ActionMove`.

szapp:
No any fight action is inhibited while falling. Falling is determined, if distance between player model and ground is too large - much like the engine does at different points.

PR #224, opened 2021-05-09
Exploit: Undead orc priest can die from fall damage

AmProsius:
**Describe the bug**
Grash-Varrag-Arushat (the last undead orc priest) can die by fall damage rather than by Uriziel damage only.

**Expected behavior**
Grash-Varrag-Arushat (the last undead orc priest) can now only be killed with Uriziel.

**Additional context**
Bug provided by [Blubbler](https://forum.worldofplayers.de/forum/threads/1574630-Release-Gothic-1-Community-Patch/page2?p=26716794&viewfull=1#post26716794).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Fix [#55](https://g1cp.org/issues/55) (updated): Grim now correctly mentions In Extremo in the second chapter even if the concert has not yet started playing. For details on the fix, see v1.1.0.
 * Fix [#129](https://g1cp.org/issues/129): Drake's body skin color now matches his head.
 * Fix [#220](https://g1cp.org/issues/220): Gor Na Ran no longer attacks the player character in chapter 6.
+* Fix [#224](https://g1cp.org/issues/224): Grash-Varrag-Arushat (the last undead orc priest) can now only be killed with Uriziel.
 * Fix [#226](https://g1cp.org/issues/226): A misplaced mana potion is now correctly inserted in one of the chests in the crypt under the stonehenge.
 
 ## [v1.1.0](https://g1cp.org/releases/tag/v1.1.0) (2021-05-01)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Fix [#55](https://g1cp.org/issues/55) (updated): Grim now correctly mentions In Extremo in the second chapter even if the concert has not yet started playing. For details on the fix, see v1.1.0.
 * Fix [#129](https://g1cp.org/issues/129): Drake's body skin color now matches his head.
 * Fix [#220](https://g1cp.org/issues/220): Gor Na Ran no longer attacks the player character in chapter 6.
-* Fix [#224](https://g1cp.org/issues/224): Grash-Varrag-Arushat (the last undead orc priest) can now only be killed with Uriziel.
+* Fix [#224](https://g1cp.org/issues/224): Grash-Varrag-Arushat (the last undead orc priest) can no longer be killed by fall damage.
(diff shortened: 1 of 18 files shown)


szapp:
> I wasn't able to reproduce the test case yet, because I lose focus before the undead orc hits the ground.

I think the trick of this exploit is to hold CTRL or LMB before the NPC falls, to maintain the focus.

AmProsius:
I did that, but after a few moments of falling, I lose focus nonetheless. I can give you a savegame if you want to test it yourself.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Fix [#55](https://g1cp.org/issues/55) (updated): Grim now correctly mentions In Extremo in the second chapter even if the concert has not yet started playing. For details on the fix, see v1.1.0.
 * Fix [#129](https://g1cp.org/issues/129): Drake's body skin color now matches his head.
 * Fix [#220](https://g1cp.org/issues/220): Gor Na Ran no longer attacks the player character in chapter 6.
+* Fix [#224](https://g1cp.org/issues/224): Grash-Varrag-Arushat (the last undead orc priest) can now only be killed with Uriziel.
```
szapp, line 12: How about: "... can no longer be killed by fall damage." because the fix does not actually ensure the killing with Uriziel.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix224_OrcPriestFallDamage.d
```diff
@@ -0,0 +1,55 @@
...
+    if (!Hlp_IsValidNpc(npc)) {
+        return FALSE;
+    };
+    
+    if (npc.fallDownDamage != 10) {
+        return FALSE;
+    };    
+
+    npc.fallDownDamage = 0;
+    return TRUE;
```
szapp, line 31: Although I understand the argument of avoiding too deeply nested if-blocks, I really feel that it is, on the one hand not necessary (it's just one nested if-block), and on the other hand creates three exit points of the function. Most importantly it creates much more code than necessary (and writing such a fix more work), and makes a concise and easily readable function much more complex and more difficult to follow at first glance.

The [previous flow of reading](https://g1cp.org/issues/224#implementation-draft) was "Set fallDownDamage to zero if all conditions are met", i.e. only do something if certain conditions are **true**. Now it is "If the Npc is ***in**valid*, return. If the fallDownDamage is ***not*** zero, return. Set the fallDownDamage to zero." The logical and semantic flow is lost and it takes a few moments to put these separated if-blocks together. By the multiple exit points of the functions it is harder to spot when there is actually something happening.

Semantically, I agree that "exiting early, when expected conditions are not met" makes a lot of sense for *test functions* (as they are currently mostly implemented). But for fix functions, it makes more sense the other way around: "fix something if all conditions are met". Only in the off-chance of stacking very, very many conditions or checking something general, e.g. if symbols exist, then it makes sense.

szapp (commented):
@AmProsius some thoughts on exiting-early for this particular case and fix functions in general.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix224_OrcPriestFallDamage.d
AmProsius: The more important aspect for me is that the primary action is at the top level of the function, not "hidden" inside nested conditions. But I will most certainly think about this.

szapp:
I refactored the fix, because it did not work when the players loads a game with the orc priest out of spawn range. The new implementation simplifies the fix and turns it into a simple session fix (the fall damage is not written to the game save, so does not need to be reverted).

The fall damage is now set during the dialog conditions of the orc priest. That makes the fix more robust irrespective of loading and spawning.

Additionally, I improved the test to teleport the player directly to a suitable condition to trigger the dialog and make the orc priest fall.

PR #225, opened 2021-05-11
Mordrag can be told to "get out" in New Camp

AmProsius:
**Describe the bug**
Mordrag can be told to "get out of here" even if he guided the player to the New Camp.

**Expected behavior**
Mordrag can no longer be told to "get out of here" if he guided the player to the New Camp.

**Additional context**
Bug provided by [Blubbler](https://forum.worldofplayers.de/forum/threads/1574630-Release-Gothic-1-Community-Patch/page2?p=26716794&viewfull=1#post26716794).

Reference for the proposed fix:

https://github.com/AmProsius/gothic-1-community-patch/blob/e66b303623c042ec5cfc387a1b0d4bd352605f12/scriptbase/_work/Data/Scripts/Content/Story/Missions/DIA_ORG_826_Mordrag.d#L328-L334

### DIFF
--- a/docs/changelog_de.md
+++ b/docs/changelog_de.md
@@ -15,6 +15,7 @@
 * Fix [#129](https://g1cp.org/issues/129): Drakes Hautfarbe des Körpers entspricht jetzt der seines Kopfes.
 * Fix [#220](https://g1cp.org/issues/220): Gor Na Ran no longer attacks the player character in chapter 6.
 * Fix [#224](https://g1cp.org/issues/224): Grash-Varrag-Arushat (der letzte untote Ork-Priester) kann nicht mehr durch Fallschaden getötet werden.
+* Fix [#225](https://g1cp.org/issues/225): Mordrag kann nicht mehr gesagt werden "Mach, dass du wegkommst", wenn er den Spieler-Charakter bereits ins Neue Lager geführt hat.
 * Fix [#226](https://g1cp.org/issues/226): Ein falsch platzierter Manatrank ist nun korrekt in einer der Truhen in der Gruft unter dem Stonehenge aufzufinden.
 * Fix [#228](https://g1cp.org/issues/228): Nach Gorn kann nun korrekt in Ambient-Dialogen gefragt werden, nachdem der Spieler mit Lares gesprochen hat.
 * Fix [#234](https://g1cp.org/issues/234): Im Tagebucheintrag für Drax als Lehrer "Wissen über Zähne entfernen - Wolf, Orchund, Snapper, Beisser, Bluthund, Schattenläufer." korrigiert zu "Wissen über Zähne entfernen - Wolf, Orkhund, Snapper, Beißer, Bluthund, Schattenläufer.".
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix225_MordragNcGetOut.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix225_MordragNcGetOut.d
@@ -1,0 +1,48 @@
+/*
+ * #225 Mordrag can be told to "get out" in New Camp
+ */
+func int G1CP_225_MordragNcGetOut() {
+    const string CONDITION = ""; CONDITION = "Org_826_Mordrag_HauAb_Condition";
+
+    if (!G1CP_IsFunc(CONDITION, "int|none"))
+    || (!G1CP_IsInfoInst("Org_826_Mordrag_GotoNewcamp"))
+    || (!G1CP_IsIntVar("Thorus_MordragKo", 0)) {
+        return FALSE;
+    };
+
+    HookDaedalusFuncS(CONDITION, "G1CP_225_MordragNcGetOut_Hook");
+    return TRUE;
+};
+
+/*
+ * This function intercepts the dialog condition to introduce more conditions
+ */
+func int G1CP_225_MordragNcGetOut_Hook() {
+    G1CP_ReportFuncToSpy();
+
+    // Define possibly missing symbols locally
+    const int LOG_RUNNING = 1;
+
+    // Add the new conditions (other conditions remain untouched)
+    var int cond1;
+    var int cond2;
+
+    // Check if dialog was told (check if symbol exists first!)
+    var int symbId; symbId = MEM_GetSymbolIndex("Org_826_Mordrag_GotoNewcamp");
+    if (symbId != -1) {
+        cond1 = !Npc_KnowsInfo(hero, symbId);
+    } else {
+        cond1 = TRUE; // If the dialog does not even exist take no chances
+    };
+
+    // Check if quest is running (If that variable does not even exist take no chances)
+    cond2 = (G1CP_GetIntVar("Thorus_MordragKo", 0, LOG_RUNNING) == LOG_RUNNING);
+
+    // Return false if either of the conditions is false
+    if ((!cond1) || (!cond2)) {
+        return FALSE;
+    };
+
+    // Continue with the original function
+    ContinueCall();
+};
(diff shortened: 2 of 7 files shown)


docs/changelog_de.md
```diff
@@ -15,6 +15,7 @@
 * Fix [#129](https://g1cp.org/issues/129): Drakes Hautfarbe des Körpers entspricht jetzt der seines Kopfes.
 * Fix [#220](https://g1cp.org/issues/220): Gor Na Ran no longer attacks the player character in chapter 6.
 * Fix [#224](https://g1cp.org/issues/224): Grash-Varrag-Arushat (der letzte untote Ork-Priester) kann nicht mehr durch Fallschaden getötet werden.
+* Fix [#225](https://g1cp.org/issues/225): Mordrag kann nicht mehr gesagt werden "Mach, dass du wegkommst", wenn er den Spieler-Charakter bereits ins Neue Lager geführt hat.
```
szapp, line 18: I would suggest to just write "Spieler" here, not "Spieler-Character", since the player themself is actually guided. In the English version of this changelog entry it is also just "player".

src/Ninja/G1CP/Content/Fixes/Session/fix225_MordragNcGetOut.d
```diff
@@ -0,0 +1,48 @@
+/*
+ * #225 Mordrag can be told to "get out" in New Camp
+ */
+func int G1CP_225_MordragNcGetOut() {
+    const string CONDITION = ""; CONDITION = "Org_826_Mordrag_HauAb_Condition";
+
+    if (!G1CP_IsFunc(CONDITION, "int|none"))
```
szapp, line 7: I am all for avoiding redundancy and all, but don't you think that is a bit overkill. It obscures readability and makes the function longer and more work to type up simple fixes, while not really helping much for redundancy. Aside from that, the syntax is really weird here (why not define the constant directly), and more confusing then helping. I strongly advise to just write out the string in both occurrences, as done everywhere else in the scripts of similar fixes.

src/Ninja/G1CP/Content/Fixes/Session/fix225_MordragNcGetOut.d
```diff
@@ -0,0 +1,48 @@
...
+    var int cond1;
+    var int cond2;
+
+    // Check if dialog was told (check if symbol exists first!)
+    var int symbId; symbId = MEM_GetSymbolIndex("Org_826_Mordrag_GotoNewcamp");
+    if (symbId != -1) {
+        cond1 = !Npc_KnowsInfo(hero, symbId);
+    } else {
+        cond1 = TRUE; // If the dialog does not even exist take no chances
+    };
```
szapp, line 36: This construct is not necessary (may have been some left over in a similar fix). The existence of "Org_826_Mordrag_GotoNewcamp" is already guaranteed by the function above. If it does not exist, this function here would never be executed. This one line here should suffice:
```D
// Check if dialog was told
cond1 = (!Npc_KnowsInfo(hero, MEM_GetSymbolIndex("Org_826_Mordrag_GotoNewcamp")));
```

src/Ninja/G1CP/Content/Fixes/Session/fix225_MordragNcGetOut.d
```diff
@@ -0,0 +1,48 @@
...
+
+    // Check if quest is running (If that variable does not even exist take no chances)
+    cond2 = (G1CP_GetIntVar("Thorus_MordragKo", 0, LOG_RUNNING) == LOG_RUNNING);
+
+    // Return false if either of the conditions is false
+    if ((!cond1) || (!cond2)) {
+        return FALSE;
+    };
```
szapp, line 44: I would turn these conditions around. There are some double negations here and it is getting quite confusing to follow the logic of this fix. I suggest to do the following:
```D
// Check if dialog was told
cond1 = Npc_KnowsInfo(hero, MEM_GetSymbolIndex("Org_826_Mordrag_GotoNewcamp"));

// Check if quest is not running
cond2 = (G1CP_GetIntVar("Thorus_MordragKo", 0, LOG_RUNNING) != LOG_RUNNING);

// Return false if either condition is met
if (cond1) || (cond2) {
    return FALSE;
};
```

src/Ninja/G1CP/Content/Tests/test225.d
```diff
@@ -0,0 +1,42 @@
...
+
+    // Backup values
+    var int toldBak; toldBak = Npc_KnowsInfo(hero, infoId);
+    var int aiVarBak; aiVarBak = G1CP_NpcGetAiVarI(npc, aiVarId, FALSE);
+    var int varBak; varBak = G1CP_GetIntVarI(varId, 0, 0);
+
+    // Set new values
+    G1CP_SetInfoToldI(infoId, TRUE);
+    G1CP_NpcSetAiVarI(npc, aiVarId, TRUE);
+    G1CP_SetIntVarI(varId, 0, 0);
```
szapp, line 24: Zero is semantically not a valid value for a quest indicator variable. It should be set to either `LOG_RUNNING`, `LOG_SUCCESS`, or similar.

src/Ninja/G1CP/Content/Tests/test225.d
```diff
@@ -0,0 +1,42 @@
...
+
+    // Call dialog condition function
+    G1CP_Testsuite_Call(funcId, npc, hero, FALSE);
+    var int ret; ret = MEM_PopIntResult();
```
szapp, line 28: To cover all ends, this test is incomplete. The fix introduces *two* new or-conditions for which the dialog condition should fail, but the dialog condition is only tested here *once*, for both of them at the same time. The test should have two passes (i.e. calling the dialog condition twice), for either of the two conditions. Only that will fully cover testing the fix.

szapp (commented):
@AmProsius this will need some fixing.

AmProsius:
Okay, thanks for the review! I based this fix and test on #153 (fix of #33), so I guess I'll have to update that fix and test as well.

szapp:
Good catch. The fix of #33 is slightly different. Here we have two independent and-conditions (if ***either*** is false return false), in #33 we have two or-conditions (if ***both*** are false return false), compare [this](https://github.com/AmProsius/gothic-1-community-patch/pull/225#issuecomment-813089761) with [this](https://github.com/AmProsius/gothic-1-community-patch/pull/153#issuecomment-790109718).

It may not hurt to negate the conditions in #33, but also does not seem necessary (there are no double negations). The symbol check in #33 is important and should remain. Unlike here, the existence of the symbol has to be checked again, because it is not guaranteed that it exists:
https://github.com/AmProsius/gothic-1-community-patch/blob/f386ca443496783307a91437f00772b92802a4dd/src/Ninja/G1CP/Content/Fixes/Session/fix033_ShrikeQuestDialog.d#L6

As for the test of #33, it does not need two passes, because there are or-conditions. However, this line should be adjusted as mentioned above:
https://github.com/AmProsius/gothic-1-community-patch/blob/f386ca443496783307a91437f00772b92802a4dd/src/Ninja/G1CP/Content/Tests/test033.d#L24

src/Ninja/G1CP/Content/Tests/test225.d
AmProsius: The fix introduces two new **and**-conditions, so I don't see the problem here. Or am I mistaken?

szapp: Yes, you are right. The dialog condition returns true if both new conditions are met, but the dialog condition returns false if either of the new conditions if false. The test currently confirms that the dialog condition returns false if the dialog "Org_826_Mordrag_GotoNewcamp" was told. What the test should also check is, that the dialog condition also returns false if only the variable "Thorus_MordragKo" is not set to `LOG_RUNNING`.

AmProsius: Ah I see. Thanks for the explanation, I will update the test accordingly.

szapp: Thanks!

Issue #42, opened 2021-01-06
Guards have duplicated END dialog 


AmProsius:
Two Guards have duplicated END dialogs.

catalinstoian:
DIA_GRD_218_GuardPassage.d

```
/*
instance DIA_Grd_218_Exit (C_INFO)
{
	npc			= Grd_218_GArdist;
	nr			= 999;
	condition	= DIA_Grd_218_Exit_Condition;
	information	= DIA_Grd_218_Exit_Info;
	permanent	= 1;
	description = DIALOG_ENDE;
};                       

FUNC int DIA_Grd_218_Exit_Condition()
{
	return 1;
};

FUNC VOID DIA_Grd_218_Exit_Info()
{
	AI_StopProcessInfos	(self);
};
*/
```

DIA_GRD_245_GuardPassage.d

```
/*
instance DIA_GRD_245_Exit (C_INFO)
{
	npc			= GRD_245_Gardist;
	nr			= 999;
	condition	= DIA_GRD_245_Exit_Condition;
	information	= DIA_GRD_245_Exit_Info;
	permanent	= 1;
	description = DIALOG_ENDE;
};                       

FUNC int DIA_GRD_245_Exit_Condition()
{
	return 1;
};

FUNC VOID DIA_GRD_245_Exit_Info()
{
	AI_StopProcessInfos	(self);
};
*/
```

AmProsius:
This issue was split into two separate issues (see #244).

PR #128, opened 2021-07-17
Guard has duplicated END dialog

AmProsius:
**Describe the bug**
The guard right next to Glen in the old mine has two ending dialogue options. 

**Expected behavior**
There should be just one option to end the dialogue.

**Additional context**
Instance `Grd_264_Gardist`.


szapp (changes requested):
I think it's quite elegant to piggyback #42!

However, the fix status of #42 will mark as failed if the two dialogs in #42 are in fact fixed but the one from #128 is not. Same for the opposite. That is highly misleading and could cause a lot of headaches during testing.

I propose to rather copy and adjust the function `G1CP_042_GuardExitDialog` while re-using the function `G1CP_042_ConfirmByteCode` (no need to copy that one).

AmProsius:
Wouldn't it be even better then to split #42 in two separate fixes?

szapp:
> Wouldn't it be even better then to split the original #42 in two separate fixes as well?

Yes, I agree. This would allow for best inspection during testing. But I fear that it would convolute the fix list / changelog, with many identical entries. An alternative is to collect all of them in one fix, as done in #42 so far.

Both options have their pros and cons, I will leave this decision up to you.

szapp:
Thanks for taking care of it. There are a few things that may not work with the new changes. I don't have the time to get into it right now, but hold out before merging this PR.

szapp (changes requested):

szapp (approved):

Issue #79, opened 2021-01-25
Wolf teaches guards dexterity instead of bandits/mercs


i5sue5:
When you've become a guard Wolf will start to teach you dexterity. This doesn't make any sense since he's a member of the new camp.

i5sue5:
---------------------------------
_work/data/Scripts/content/Story/Missions/DIA_ORG_855_Wolf.d:
-----------------------------------                

FUNC INT ORG_855_Wolf_Teach_Condition()
{
	if (C_NpcBelongsToNewCamp (other))
	{
		return TRUE;
	};
};

PR #32, opened 2021-03-05
Gorn attacks the player in the Free Mine

szapp:
**Describe the bug**
Gorn sometimes attacks the player on the raid of the Free Mine.

**Expected behavior**
Gorn no longer attacks the player on the raid of the Free Mine.

**Test**
Run manual test with `test 32`.

- [x] Gorn no longer scolds the player for killing NPCs of the Old Camp when talking to him.
      *Triggered by killing a guard when out of sight of Gorn.*
- [x] Gorn no longer warns the player of attacking him again when talking to him.
      *Triggered by defeating guards of the Old Camp.*
- [x] Gorn no longer attacks the player.
      *Triggered by Gorn seeing a guard of the Old Camp attacking the PC first.*

PR #168, opened 2021-07-31
Gor Na Bar offers to become a templar to anyone

AmProsius:
**Describe the bug**
No matter the player's guild, Gor Na Bar tells him to speak to Cor Angar about becoming a templar and a log entry is created. 

**Expected behavior**
Gor Na Bar now tells only novices to speak to Cor Angar about becoming a templar.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix168_GorNaBarBecomeTemplar.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix168_GorNaBarBecomeTemplar.d
@@ -20,6 +20,13 @@ func int G1CP_168_GorNaBarBecomeTemplar_Hook() {
     // Define possibly missing symbols locally
     const int GIL_NOV = 12;
 
+    // Update the symbols given they exist (once)
+    const int updated = FALSE;
+    if (!updated) {
+        GIL_NOV = G1CP_GetIntConst("GIL_NOV", 0, GIL_NOV);
+        updated = TRUE;
+    };
+
     // Add the new condition (other conditions remain untouched)
     if (Npc_GetTrueGuild(hero) != GIL_NOV) {
         return FALSE;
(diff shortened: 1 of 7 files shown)


src/Ninja/G1CP/Content/Fixes/Session/fix168_GorNaBarBecomeTemplar.d
```diff
@@ -0,0 +1,30 @@
...
+};
+
+/*
+ * This function intercepts the dialog condition to introduce more conditions
+ */
+func int G1CP_168_GorNaBarBecomeTemplar_Hook() {
+    G1CP_ReportFuncToSpy();
+
+    // Define possibly missing symbols locally
+    const int GIL_NOV = 12;
```
szapp, line 21: I wonder if the constant should rather be retrieved here. The value (12) may have been changed in the scripts. How have we done this in similar fixes?

AmProsius: From what I could find we always defined the guild constants.

szapp: Ah. I wonder what the idea was there and if I am missing something now. Do you happen do have a (few) examples at hand at the moment? I couldn't find one so fast.

szapp: Ignoring the test suite, I found that we retrieved the guild in the fixes of #79 and also #32. I could not find any other fixes regarding guilds.

szapp: We can leave the above line untouched but add a [block like found here](https://github.com/AmProsius/gothic-1-community-patch/blob/1d4b915c0f50c75e0b20eb0971907cc62b1929c7/src/Ninja/G1CP/Content/Fixes/Session/fix032_GornAttackFreeMine.d#L87-L96).

```D
    // Update the symbols given they exist (once)
    const int updated = FALSE;
    if (!updated) {
        GIL_NOV = G1CP_GetIntConst("GIL_NOV", 0, GIL_NOV);
        updated = TRUE;
    };
```

I propose to use this once-block here, because the condition function is called multiple times. Retrieving the constant only once may be beneficial to performance (although the impact will not be noticeable).

PR #57, opened 2021-08-12
Misplaced texture near Gravo 

AmProsius:
There's a misplaced texture near Gravo at the water.


szapp:
@AmProsius This PR should be done too. Here, too, I have not tested whether the texture is back at its original positing when unloading the patch.

AmProsius:
When starting the game (again) without the patch, the flickering texture is back.

Issue #16, opened 2020-12-30
Thorus' bribe dialog doesn't disappear


AmProsius:
Thorus can be bribed if the player has already obtained the permit to an audience with Gomez.
The option to bribe Thorus stays open after the player has already bribed him.

catalinstoian:
DIA_GRD_200_Thorus.d

// ************************************************************
// 						Give1000Ore					
// ************************************************************
var int thorus_bribed;

INSTANCE Info_Thorus_Give1000Ore (C_INFO)
{
	npc			= GRD_200_THORUS;
	nr			= 1;
	condition	= Info_Thorus_Give1000Ore_Condition;
	information	= Info_Thorus_Give1000Ore_Info;
	permanent	= 1;
	description = "I have the 1000 nuggets of ore. Now let me go in!";
};                       

FUNC INT Info_Thorus_Give1000Ore_Condition()
{	
	if (Npc_KnowsInfo(hero,Info_Thorus_BribeGuard) && (Diego_GomezAudience == FALSE) && (thorus_bribed == FALSE))
	{
		return 1;
	};
};

FUNC VOID Info_Thorus_Give1000Ore_Info()
{
	AI_Output (other, self,"Info_Thorus_Give1000Ore_15_00"); //I have the 1000 nuggets of ore. Now let me go in!
	
	if (Npc_HasItems(other, ItMiNugget)>=1000)
	{
		B_GiveInvItems 	(other,self,ItMiNugget,1000);		
		AI_Output (self, other,"Info_Thorus_Give1000Ore_09_01"); //Alright, go ahead. You can go in the castle, but don't do anything stupid, okay?
		var C_NPC wache212; wache212 = Hlp_GetNpc(Grd_212_Torwache);
		var C_NPC wache213; wache213 = Hlp_GetNpc(Grd_213_Torwache);
		wache212.aivar[AIV_PASSGATE] = TRUE;
		wache213.aivar[AIV_PASSGATE] = TRUE;
		thorus_bribed = TRUE;
	}
	else
	{
		AI_Output (self, other,"Info_Thorus_Give1000Ore_09_02"); //Don't try'n trick me, kid. You don't have 1000 nuggets of ore!
	};
};

// ************************************************************
// 					Brief für Magier
// ************************************************************

INSTANCE Info_Thorus_LetterForMages (C_INFO)
{
	npc			= GRD_200_THORUS;
	nr			= 1;
[...]

PR #166, opened 2021-10-10
Ambient dialogs of rogue not available

AmProsius:
**Describe the bug**
Two ambient dialogs of a rogue in the New Camp are not available due to a missing variable set.

**Expected behavior**
Ambient dialogs of a rogue in the New Camp are now available.

**Additional context**
https://github.com/AmProsius/gothic-1-community-patch/blob/0623f55a4ba7811367e2bc52a879724c4616717e/scriptbase/_work/Data/Scripts/Content/Story/Missions/DIA_ORG_829_Organisator.d#L111-L117

https://github.com/AmProsius/gothic-1-community-patch/blob/0623f55a4ba7811367e2bc52a879724c4616717e/scriptbase/_work/Data/Scripts/Content/Story/Missions/DIA_ORG_829_Organisator.d#L141-L147

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix166_RogueWeedDialog.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix166_RogueWeedDialog.d
@@ -1,0 +1,94 @@
+/*
+ * #166 Ambient dialogs of rogue not available
+ */
+func int G1CP_166_RogueWeedDialog() {
+    if (!G1CP_IsFunc("Info_ORG_829_SpecialInfo_Condition", "int|none"))
+    || (!G1CP_IsIntVar("Org_829_GotJoint", 0))
+    || (!G1CP_IsInfoInst("Info_ORG_829_OfferJoint"))
+    || (!G1CP_IsItemInst("ItMiJoint_1"))
+    || (!G1CP_IsItemInst("ItMiJoint_2"))
+    || (!G1CP_IsItemInst("ItMiJoint_3")) {
+        return FALSE;
+    };
+
+    HookDaedalusFuncS("Info_ORG_829_SpecialInfo_Condition", "G1CP_166_RogueWeedDialog_Hook1");
+    HookDaedalusFuncS("Info_ORG_829_PERM_Condition", "G1CP_166_RogueWeedDialog_Hook2");
+    return TRUE;
+};
+
+/*
+ * This function intercepts the dialog condition to introduce more conditions
+ */
+func int G1CP_166_RogueWeedDialog_Hook1() {
+    G1CP_ReportFuncToSpy();
+
+    // Symbol indices (existence confirmed by function above)
+    var int joint1Id; joint1Id = MEM_GetSymbolIndex("ItMiJoint_1");
+    var int joint2Id; joint2Id = MEM_GetSymbolIndex("ItMiJoint_2");
+    var int joint3Id; joint3Id = MEM_GetSymbolIndex("ItMiJoint_3");
+    var int gotJointId; gotJointId = MEM_GetSymbolIndex("Org_829_GotJoint");
+
+    // Add the new conditions (other conditions remain untouched)
+    var int cond1;
+    var int cond2;
+    var int cond3;
+    var int cond4;
+
+    // Check if dialog was told (check if symbol exists first!)
+    cond1 = !Npc_KnowsInfo(hero, MEM_GetSymbolIndex("Info_ORG_829_OfferJoint"));
+
+    // Check if NPC has a joint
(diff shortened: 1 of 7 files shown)


src/Ninja/G1CP/Content/Fixes/Session/fix166_RogueWeedDialog.d
```diff
@@ -0,0 +1,94 @@
...
+
+    // Set the variable
+    G1CP_SetIntVarI(gotJointId, 0, TRUE);
+
+    // Continue with the original function
+    ContinueCall();
+};
+
+/*
+ * Exact copy of the function above. Need unique functions for both because of the way Daedalus hooks work
```
szapp, line 58: I wonder, instead of having two near identical functions, we could have the hook functions call a common one. This "super function" would return true or false. Depending on the return value the hook functions would then either call `ContinueCall` or return false themselves.

src/Ninja/G1CP/Content/Fixes/Session/fix166_RogueWeedDialog.d
```diff
@@ -0,0 +1,94 @@
...
+    // Check if dialog was told (check if symbol exists first!)
+    cond1 = !Npc_KnowsInfo(hero, MEM_GetSymbolIndex("Info_ORG_829_OfferJoint"));
+
+    // Check if NPC has a joint
+    cond2 = !Npc_HasItems(self, joint1Id);
+    cond3 = !Npc_HasItems(self, joint2Id);
+    cond4 = !Npc_HasItems(self, joint3Id);
+
+    // Return false if either of the conditions is true
+    if (cond1) || (cond2 && cond3 && cond4) {
```
szapp, line 46: This if-condition might cause issues. It will prevent the dialog from appearing if the variable is `Org_829_GotJoint` is true but the NPC no longer has the items. We will have to revise this if-condition. We can brain storm some ideas here.

Since we are adding an or-condition, it is dangerous to prematurely return false. Likewise, it is dangerous to return true, because a mod could have added new conditions in the dialog condition function.

I wonder if it will be enough to just change the order: Run the original function first (use `ContinueCall()` in the beginning and collect its return value). If the return value is true, return true, if it is false, add our new conditions and return true/false. If I am not mistaken we might have done this in a few fixes. Maybe we can find those dialog fixes.

src/Ninja/G1CP/Content/Tests/test166.d
```diff
@@ -0,0 +1,29 @@
+/*
+ * #166 Ambient dialogs of rogue not available
+ *
+ * There does not seem an easy way to test this fix programmatically, so this test relies on manual confirmation.
```
szapp, line 4: There should actually be a way to check this with an automatic test. As found in similar tests, set the dialog `Info_ORG_829_OfferJoint` to told, trigger the condition function, and check its return value.

szapp (commented):
I know this fix already found its way into the release, but we should revise it. I didn't have the time to review it until now. We can discuss how to improve it here.

src/Ninja/G1CP/Content/Fixes/Session/fix166_RogueWeedDialog.d
AmProsius: Actually this is how we handles these types of issues in the past (see #16 for reference). We should revisit these fixes as well then.

szapp: I agree with the comment about #66. That may be worth worrying about, if the dialog was an important one and is likely to be changed in a mod. Here - you are right - there is no need to overthink it. 

Nevertheless, the issue I mentioned in the first paragraph in the comment above should be addressed. And I think your suggestion with #96 sounds great. 

What we should do is check if the variable that is used to check whether the joint has been given is false. If so, check if the dialog was told and the NPC has either of the items and then set the variable to true. After that (irrespective of the check of the variable) the original function should be triggered.

kot9ka:
Why? Maybe it's a feature by devs. Guards take a joint and 
and they are silently happy. 
You didn't think about it?

szapp:
> Why? Maybe it's a feature by devs. Guards take a joint and 
> 
> and they are silently happy. 
> 
> You didn't think about it?

It is, of course, always a possibility that the developers intended a certain circumstance. Thank you for voicing this concern.

However, it is rather unlikely here, because the existence of the dialogs and conditions and variables strongly suggest an oversight. If it were intentional or a last-minute change, this circumstance would have most likely been implemented differently. This hunch based on other examples of last-minute changes and intended omissions we found in the scripts.

PR #289, opened 2022-01-12
Check requirements during setup

szapp:
To avoid confusion among some players it may be beneficial to at least inform the player of the installation requirements during the setup. It seems to happen occasionally that players are not aware of the requirements and wonder why the G1CP has no effect on their game.

This is already facilitated well by Spine (as it checks requirements on game starts), but not with the manual installation setup we provide here.

The setup could plainly inform about all requirements in the finish page of the setup wizard, or optionally check if they are already met and only inform about the missing ones.

Things to check (if possible with the NSIS setup):
- [x] **Gothic on version 1.08k**  
NSIS would either have to read the file properties ~~or confirm the checksum of the EXE-file. That might fail, because some third-party tools may modify the EXE file.~~
- [x] **Ninja installed**  
Merely check for the existence of ./System/Ninja.dll.
- [x] **Minimally required Ninja version**  
This would require to read the file properties of the DLL-file. ~~Not sure if NSIS is capable of that.~~
- [x] **Data execution prevention disabled**  
~~This is most likely not detectable via NSIS.~~
- [x] *Optional* **Check for SystemPack or Union**  
This should only be advertised as a recommendation. To check for them, see the setup source of Ninja.
- [x] *Optional* **Gothic is not installed in the Program Files directories**
This is quite easy to check.


szapp:
@AmProsius Could you try the changes to the setup, with `make release`.

After choosing the installation directory (only selectable if it contains `System\Gothic.exe` as before), and then pressing the "Install" button, the setup will now perform the checks as mentioned in the PR description (see above).

If any of them is not met, a message box will educate about the missing requirement. The setup can still be forced to continue, with a note about the G1CP not working in that current state. The setup should even be able to detect if the Data Execution Prevention feature is turned on.

If you find the time, let me know if you spot any issues.

szapp:
To build the setup, the NSIS Plugin “MoreInfo” needs to be installed. The involves copying the MoreInfo.dll into the respective Plugin directory inside the local NSIS installation.

Download: http://forums.winamp.com/attachment.php?s=a2e2da95772d82867dffe26f1a55ae59&attachmentid=48651&d=1301181534

AmProsius:
Would it be possible and practical to include Ninja in the setup for those who haven't installed it already?

szapp:
> Would it be possible and practical to include Ninja in the setup for those who haven't installed it already?

No. There may be new versions of Ninja coming. At most links to the mentioned installation requirements may be handy.

With the current implementation, there are only simple message boxes that cannot show clickable links well. Alternatively, we could create a page in the setup (that only is visible should some requirements not be met) to list the ones that are missing including a link. However, I found a message box with a clear yes and no, more disrupting than simply clicking through a setup wizard without reading anything of the generic texts.

szapp:
Actually, a separate page to list the non-met installation requirements might be quite neat. The "install" button (to force installation anyway) in the bottom of the wizard could be grayed out and made clickable only by checking a checkbox stating that the user acknowledges that the G1CP will not run in the current configuration.

AmProsius:
> Actually, a separate page to list the non-met installation requirements might be quite neat. The "install" button (to force installation anyway) in the bottom of the wizard could be grayed out and made clickable only by checking a checkbox stating that the user acknowledges that the G1CP will not run in the current configuration.

After seeing the new installer in action, I would actually prefer this solution. An optional separate page is less intrusive than an alert.

szapp:
> After seeing the new installer in action, I would actually prefer this solution. An optional separate page is less intrusive than an alert.

I will try to get on that on the weekend. My thought is to list all outstanding requirements as bullet points with a clickable link underneath each of them. The next/install button will be grayed out and the back button will be in focus. The next/install button can still be activated to force the installation anyway, by checking a checkbox underneath the bullet points. Does this sound good?

szapp:
I added a verification page that shows more details about the non-met installation requirements. The layout (i.e. positioning of the window controls) is outsourced and easily adjustable from the file `./setup/checks.ini`.

Another thing to note is that the setup no longer checks for the file `GothicMod.exe` but only for `Gothic.exe`. In a vanilla Gothic installation, there is no `GothicMod.exe`. What I cannot remember is, whether this file is added through installing the Playerkit or also by simply installing the 1.08k patch. I cannot check this myself because I don't have my Gothic installation CDs with me. Do you have the means to confirm this, @AmProsius? It is not a pressing issue.

There are a few different ways, the installation requirements will not be met. I have listed all of them here. Annotations correspond to the screenshots below.

@AmProsius It would of course be great, if you can confirm these cases on a second computer, to make sure there are no severe bugs in the setup.

1. Gothic is not patched to version 1.08k.<sup>[1]</sup>
2. Ninja is not installed.<sup>[1]</sup>
3. The version of the installed Ninja cannot be read, i.e. the installation is invalid or corrupt.<sup>[2]</sup>
4. The version of Ninja is below the minimum requirement for the G1CP.<sup>[3]</sup>
5. DEP is globally enabled ("AlwaysOn").<sup>[1][5]</sup>
6. DEP is enabled for all processes with listed exceptions. *Note* The setup is not able to detect whether Gothic is on the list of exceptions. Therefore, this case is treated like a mere warning, and the message about below the checkbox is not visible.<sup>[6]</sup>.
7. Neither SystemPack and Union are not installed. *Note* This case is also only treated as a warning, because these tools are only recommendations.<sup>[1][4]</sup>

### English
[1]
g1

[2]
g2

[3]
g3

[4]
g4

[5]
g5

[6]
g6


### German
These are the same screenshots as above in German localization.  

<details><summary><b>Click to expand</b></summary>
[1]
g1

[2]
g2

[3]
g3

[4]
g4

[5]
g5

[6]
g6
</details>

szapp:
Another "requirement" that comes to mind would be to not use an Gothic installation that resides in the `Program Files` or `Program Files (x86)` directory. This is a common origin for reoccurring problems found in the world of gothic forums. This could be added as an recommendation like the SystemPack and Union. What do you think?

I also noticed that the setup requires administrative right to start. This is rather undesired for users who do not have (and do not want) administrative rights for the installation. For testing, I managed have the setup start at user privileges, however, then it crashes with a fatal error when choosing an installation location that requires administrative rights.

szapp:
I added another field to check whether the selected directory is one of the program files directories (see above). The setup now checks for that too. To complement the screenshots from above:

g7_a

g7_b

g7_c

### German
These are the same screenshots as above in German localization.  

<details><summary><b>Click to expand</b></summary>

g7_a

g7_b

g7_c

</details>

AmProsius:
> Another thing to note is that the setup no longer checks for the file `GothicMod.exe` but only for `Gothic.exe`. In a vanilla Gothic installation, there is no `GothicMod.exe`. What I cannot remember is, whether this file is added through installing the Playerkit or also by simply installing the 1.08k patch. I cannot check this myself because I don't have my Gothic installation CDs with me. Do you have the means to confirm this, @AmProsius? It is not a pressing issue.

After installing the 1.08k patch, there's still no `GothicMod.exe` present in the `system` folder.

AmProsius:
After installing the linked 1.08k patch, the issue still persists. We may have to require the playerkit. I will test that just now.

image

szapp:
Interesting. Could you send me a screenshot of the file version tab of the file properties dialog (not sure what to call it: right click properties, most right tab)? I am interested in the file version text.

AmProsius:
Of course:

image

AmProsius:
To save our latest discovery: The G1CP depends on the [https://www.worldofgothic.de/dl/download_34.htm](Playerkit), not the Path 1.08k alone.

setup/g1mod.nsi
```diff
@@ -456,3 +453,371 @@ Function PageLeaveReinstall
...
+LangString TextVerifyGothicVersion ${LANG_ENGLISH} "Please update your version."
+LangString TextVerifyNinja ${LANG_ENGLISH} "An installation of Ninja is required."
+LangString TextVerifyNinjaInvalid ${LANG_ENGLISH} "The Ninja installation is invalid or corrupted."
+LangString TextVerifyNinjaVersion ${LANG_ENGLISH} "Please update your version. Currently"
+LangString TextVerifySPUnion ${LANG_ENGLISH} "Either SystemPack or Union is highly recommended."
+LangString TextVerifyNotInProgFiles ${LANG_ENGLISH} "Gothic not in 'Program Files'"
+LangString TextVerifyBadDir ${LANG_ENGLISH} "These directories may cause issues for the game."
+LangString TextVerifyProgFilesGood ${LANG_ENGLISH} "Gothic is installed somewhere else."
+LangString TextVerifyDEPisOn ${LANG_ENGLISH} "Please disable DEP or add exemptions for Gothic."
+LangString TextVerifyDEPexclude ${LANG_ENGLISH} "Make sure Gothic is listed as exemption."
```
AmProsius, line 491: It's *exception*, not *exemption*. See L490 and L491.

AmProsius:
After playing around with the installer a little, I suggest we don't disable the install button if only recommendations aren't met.

So the checkbox would only appear if the Playerkit is not installed, Ninja is not installed/up-to-date or DEP is enabled.

I would not change the checklist though (if you know what I mean)!

szapp:
Sounds good. For easier implementation, I would suggest to just create the page with the checkbox initially checked (and hence the install button being enabled). Would that work for you?

I was debating whether to use the WoG link, but even the WoG forum release thread links the original website. For both adequate crediting of the author and for most up-to-date versions, I prefer to link to the original websites. Since it offers an English language tab, I think that is good.

AmProsius:
> Sounds good. For easier implementation, I would suggest to just create the page with the checkbox initially checked (and hence the install button being enabled). Would that work for you?
Yes, that would work for me.

> I was debating whether to use the WoG link, but even the WoG forum release thread links the original website. For both adequate crediting of the author and for most up-to-date versions, I prefer to link to the original websites. Since it offers an English language tab, I think that is good.
Agreed.

> Also, the link to the 1.08k_mod version is wrong in the WoG download description. The download description is outdated in general.
Do you mean the Playerkit? At least  the download is correct.

szapp:
Ok!

> Do you mean the Playerkit? At least the download is correct.

Yes, I mean that the download description on WoG says "Version 1.08k_mod", but the link leads to the "Version 1.08k" (not the player kit). But that shouldn't concern us, if we are linking to the original author's website.

AmProsius:
> Yes, I mean that the download description on WoG says "Version 1.08k_mod", but the link leads to the "Version 1.08k" (not the player kit). But that shouldn't concern us, if we are linking to the original author's website.

I'm confused. I don't see the description "Version 1.08k_mod" anywhere on the [download page](https://www.worldofgothic.de/dl/download_34.htm).

szapp:
Don't worry, I was still talking about the download page of Union:

> Also I suggest to use the WOG link for Union: https://www.worldofgothic.de/dl/download_651.htm.

AmProsius:
Oh I see! That shouldn't concern us too much.

szapp:
I fixed all the feedback we discussed. The version requirement is listed as "Gothic version 1.08k_mod" instead of "Playerkit" (or similar), because it is really only about the version of the executable file. This particular Gothic version is also shipped with other installers. Explicitly requiring the Playerkit may result in confusion and re-installing it may cause butchering of Gothic installations.

Let me know if there is more things to address in this PR. Some changes in other branches I have been working on rely on some changes in here.

AmProsius:
> Let me know if there is more things to address in this PR. Some changes in other branches I have been working on rely on some changes in here.

I'll check asap.

PR #361, opened 2023-01-08
Refactor info int variable fixes

AmProsius:
Refactor the fixes where an info instance int variable is changed.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix030_SilasTrade.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix030_SilasTrade.d
@@ -2,10 +2,5 @@
  * #30 Silas trades only one time
  */
 func int G1CP_030_SilasTrade() {
-    const string infoName = "DIA_Silas_Trade";
-    const string infoProp = "permanent";
-    const int originalInt = 0;
-    const int correctedInt = 1;
-
-    return (G1CP_ChangeInfoIntVar(infoName, infoProp, originalInt, correctedInt) > 0);
+    return G1CP_ChangeInfoIntVar("DIA_Silas_Trade", "permanent", 0, 1);
 };
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix030_SilasTrade.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix030_SilasTrade.d
@@ -2,6 +2,10 @@
  * #30 Silas trades only one time
  */
 func int G1CP_030_SilasTrade() {
-    var int symbId; symbId = G1CP_GetInfoInstId("DIA_Silas_Trade");
-    return (G1CP_ReplaceAssignInt(symbId, 0, "C_Info.permanent", 0, 0, 1) > 0);
+    const string infoName = "DIA_Silas_Trade";
+    const string infoProp = "permanent";
+    const int originalInt = 0;
+    const int correctedInt = 1;
+
+    return (G1CP_ChangeInfoIntVar(infoName, infoProp, originalInt, correctedInt) > 0);
 };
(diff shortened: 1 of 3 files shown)


src/Ninja/G1CP/Content/Fixes/Session/fix030_SilasTrade.d
```diff
@@ -2,6 +2,10 @@
...
  */
 func int G1CP_030_SilasTrade() {
-    var int symbId; symbId = G1CP_GetInfoInstId("DIA_Silas_Trade");
-    return (G1CP_ReplaceAssignInt(symbId, 0, "C_Info.permanent", 0, 0, 1) > 0);
+    const string infoName = "DIA_Silas_Trade";
+    const string infoProp = "permanent";
+    const int originalInt = 0;
+    const int correctedInt = 1;
+
+    return (G1CP_ChangeInfoIntVar(infoName, infoProp, originalInt, correctedInt) > 0);
```
szapp, line 10: I would prefer to remove the constants here, as they do not add much value - in terms of redundancy or readability. The line of code below is already very clear to me. This is my impression; Let's discuss!
Additionally, the `> 0` is not necessary, because the defined function returns exclusively `TRUE` or `FALSE`

```D
return G1CP_ChangeInfoIntVar("DIA_Silas_Trade", "permanent", 0, 1);
```

src/Ninja/G1CP/Content/Fixes/Session/fix126_SharkyTrade.d
```diff
@@ -2,6 +2,10 @@
...
  */
 func int G1CP_126_SharkyTrade() {
-    var int infoId; infoId = G1CP_GetInfoInstId("Org_843_Sharky_Fisk");
-    return (G1CP_ReplaceAssignInt(infoId, 0, "C_Info.trade", 0, 1, 0) > 0);
+    const string infoName = "Org_843_Sharky_Fisk";
+    const string infoProp = "trade";
+    const int originalInt = 1;
+    const int correctedInt = 0;
+
+    return (G1CP_ChangeInfoIntVar(infoName, infoProp, originalInt, correctedInt) > 0);
```
szapp, line 10: See comment above.

src/Ninja/G1CP/Content/Misc/info.d
```diff
@@ -84,3 +84,12 @@ funt int G1CP_ChangeInfoDesc(var string infoName, var string originalStr, var st
 
     return (G1CP_ReplaceAssignStr(infoId, 0, "C_Info.description", 0, originalStr, correctedStr) > 0);
 }
+
+/*
+ * Change the value of an info's integer variable
+ */
+funt int G1CP_ChangeInfoIntVar(var string infoName, var string infoProp, var int originalInt, var int correctedInt) {
```
szapp, line 91: Typo here: func instead of funt

Right now, the function contains almost the identical set of parameters of the function that it substitutes. I am not so sure about the gains here - considering it only affects two fixes so far, that are already very concise.

Depending whether there will be more such fixes, we could decide to make two functions instead to reduce the number of parameters and increase their relevance:
```D
func int G1CP_ChangeInfoPermanent(var string infoName, var int originalOn, var int correctedOn)

func int G1CP_ChangeInfoTrade(var string infoName, var int originalOn, var int correctedOn)
```
However, this I would only propose if there is an substantial amount of such fixes.

szapp (changes requested):
Let's discuss these thoughts. I am open to also move ahead and see if adjustments are in order down the line.

src/Ninja/G1CP/Content/Fixes/Session/fix030_SilasTrade.d
AmProsius: Agreed

src/Ninja/G1CP/Content/Fixes/Session/fix126_SharkyTrade.d
AmProsius: Agreed

src/Ninja/G1CP/Content/Misc/info.d
AmProsius: Thanks for the typo hint!

I understand the motivation behind splitting the functions, but like you I don't think they are necessary at the moment. Definitely worth to revisit in the future!

PR #362, opened 2023-01-08
Refactor NPC int variable fix

AmProsius:
Refactor the fix where an NPC instance int variable is changed.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix124_GateGuardId.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix124_GateGuardId.d
@@ -2,6 +2,10 @@
  * #124 Gate guard doesn't close castle gate in chapter 4
  */
 func int G1CP_124_GateGuardId() {
-    var int npcId; npcId = G1CP_GetNpcInstId("GRD_280_Gardist");
-    return (G1CP_ReplaceAssignInt(npcId, 0, "C_Npc.id", 0, 230, 280) > 0);
+    const string npcName = "GRD_280_Gardist";
+    const string npcProp = "id";
+    const int originalInt = 230;
+    const int correctedInt = 280;
+
+    return (G1CP_NpcChangeIntVar(npcName, npcProp, originalInt, correctedInt) > 0);
 };
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix124_GateGuardId.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix124_GateGuardId.d
@@ -2,10 +2,5 @@
  * #124 Gate guard doesn't close castle gate in chapter 4
  */
 func int G1CP_124_GateGuardId() {
-    const string npcName = "GRD_280_Gardist";
-    const string npcProp = "id";
-    const int originalInt = 230;
-    const int correctedInt = 280;
-
-    return (G1CP_NpcChangeIntVar(npcName, npcProp, originalInt, correctedInt) > 0);
+    return G1CP_NpcChangeIntVar("GRD_280_Gardist", "id", 230, 280);
 };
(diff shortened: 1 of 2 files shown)


src/Ninja/G1CP/Content/Misc/npc.d
```diff
@@ -219,3 +219,12 @@ func void G1CP_NpcIdSetAiVar(var int npcInstance, var string aiVarName, var int
     var C_Npc slf; slf = Hlp_GetNpc(npcInstance);
     G1CP_NpcSetAiVar(slf, aiVarName, value);
 };
+
+/*
+ * Change the value of an NPC's integer variable
+ */
+funt int G1CP_NpcChangeIntVar(var string npcName, var string npcProp, var int originalInt, var int correctedInt) {
```
szapp, line 226: Type: funt -> func

Creating this extra layer of a function call, does not makes sense as long as there is no other fix of that same type.

src/Ninja/G1CP/Content/Misc/npc.d
```diff
@@ -219,3 +219,12 @@ func void G1CP_NpcIdSetAiVar(var int npcInstance, var string aiVarName, var int
...
+    var int npcId; npcId = G1CP_GetNpcInstId(npcName);
+
+    return (G1CP_ReplaceAssignInt(npcId, 0, ConcatStrings("C_Npc.", npcProp), 0, originalInt, correctedInt) > 0);
```
szapp, line 229: Might have to double-check if the use of `ConcatStrings` causes issues here, because the result is not a constant. I do not recall if that was a prerequisite here.

src/Ninja/G1CP/Content/Fixes/Session/fix124_GateGuardId.d
```diff
@@ -2,6 +2,10 @@
...
  */
 func int G1CP_124_GateGuardId() {
-    var int npcId; npcId = G1CP_GetNpcInstId("GRD_280_Gardist");
-    return (G1CP_ReplaceAssignInt(npcId, 0, "C_Npc.id", 0, 230, 280) > 0);
+    const string npcName = "GRD_280_Gardist";
+    const string npcProp = "id";
+    const int originalInt = 230;
+    const int correctedInt = 280;
+
+    return (G1CP_NpcChangeIntVar(npcName, npcProp, originalInt, correctedInt) > 0);
```
szapp, line 10: See comments from #361.

szapp (changes requested):

src/Ninja/G1CP/Content/Misc/npc.d
AmProsius: Thanks for the typo hint!

There is at least one other fix queued that will benefit from this refactor, so I think it's worth refactoring.

src/Ninja/G1CP/Content/Misc/npc.d
AmProsius: I can't test that right now, so I would be glad if you can confirm this. If you are right, this refactoring doesn't benefit the fix templates and is obsolete.

src/Ninja/G1CP/Content/Fixes/Session/fix124_GateGuardId.d
AmProsius: Agreed.

src/Ninja/G1CP/Content/Misc/npc.d
szapp: I would say we implement this and I will have a look in February to check whether something breaks.

szapp (approved):

src/Ninja/G1CP/Content/Misc/npc.d
AmProsius: I'll leave this PR open then, because I currently don't want to break anything.

PR #363, opened 2023-01-08
Refactor rename NPC fix

AmProsius:
Refactor the fix where an NPC is renamed.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix233_DE_OrcDogName.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix233_DE_OrcDogName.d
@@ -2,6 +2,9 @@
  * #233 Spelling - Orc Dog (DE)
  */
 func int G1CP_233_DE_OrcDogName() {
-    var int npcId; npcId = G1CP_GetNpcProtoId("Mst_Default_OrcDog");
-    return (G1CP_ReplaceAssignStr(npcId, 0, "C_NPC.NAME", 0, "Orc-Hund", "Orkhund") > 0);
+    const string npcName = "Mst_Default_OrcDog";
+    const string originalStr = "Orc-Hund";
+    const string correctedStr = "Orkhund";
+
+    return (G1CP_NpcRename(npcName, originalStr, correctedStr) > 0);
 };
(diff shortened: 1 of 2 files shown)


src/Ninja/G1CP/Content/Misc/npc.d
```diff
@@ -219,3 +219,12 @@ func void G1CP_NpcIdSetAiVar(var int npcInstance, var string aiVarName, var int
     var C_Npc slf; slf = Hlp_GetNpc(npcInstance);
     G1CP_NpcSetAiVar(slf, aiVarName, value);
 };
+
+/*
+ * Rename an NPC
+ */
+funt int G1CP_NpcRename(var string npcName, var string originalStr, var string correctedStr) {
```
szapp, line 226: Typo: funt -> func


Unlike the other recent PR, this function actually changes a particular property (the NPC's name) instead of an arbitrary string or integer. This makes this function much more reasonable than the others.

Nevertheless, as mentioned in #362, this extra function is not necessary unless there are more fixes of that kind.

src/Ninja/G1CP/Content/Fixes/Session/fix233_DE_OrcDogName.d
```diff
@@ -2,6 +2,9 @@
  * #233 Spelling - Orc Dog (DE)
  */
 func int G1CP_233_DE_OrcDogName() {
-    var int npcId; npcId = G1CP_GetNpcProtoId("Mst_Default_OrcDog");
-    return (G1CP_ReplaceAssignStr(npcId, 0, "C_NPC.NAME", 0, "Orc-Hund", "Orkhund") > 0);
+    const string npcName = "Mst_Default_OrcDog";
+    const string originalStr = "Orc-Hund";
+    const string correctedStr = "Orkhund";
+
+    return (G1CP_NpcRename(npcName, originalStr, correctedStr) > 0);
```
szapp, line 9: See comments of #361.

szapp (changes requested):

src/Ninja/G1CP/Content/Misc/npc.d
AmProsius: Thanks for the typo hint!

There is at least one other fix queued that will benefit from this refactor, so I think it's worth refactoring.

src/Ninja/G1CP/Content/Fixes/Session/fix233_DE_OrcDogName.d
AmProsius: Agreed.

szapp (approved):

PR #364, opened 2023-01-08
Refactor string array constant fixes

AmProsius:
Refactor the fixes where an element of a string array constant is changed.


src/Ninja/G1CP/Content/Misc/constant.d
```diff
@@ -0,0 +1,13 @@
+/*
+ * Replace element in string array constant
+ */
+func int G1CP_ConstReplaceStr(var string strConst, var string originalStr, var string correctedStr) {
```
szapp, line 4: We might need a different function name here, as the current name is not very descriptive. The function is **searching** a string **array** constant for a particular string element and replaces it **if it exists**. I don't have name in mind but we can discuss here.

src/Ninja/G1CP/Content_G1.src
```diff
@@ -39,6 +39,7 @@ Content\Misc\vobLight.d
 Content\Misc\npc.d
 Content\Misc\ou.d
 Content\Misc\routine.d
+Content\Misc\constant.d
```
szapp, line 42: I think this function should not be in this collection of files. As it is rather low level (replacing a string in an arbitrary string array), it should be added to the file `src/Ninja/G1CP/Content/Symbols/string.d` where the other functions used reside.

src/Ninja/G1CP/Content/Fixes/Session/fix235_DE_OrcDogMagBook.d
```diff
@@ -2,11 +2,9 @@
...
-    } else {
-        return FALSE;
-    };
-};
\ No newline at end of file
+    const string strConst = "TXT_SPELLS";
+    const string originalStr = "Verwandlung Orc-Hund";
+    const string correctedStr = "Verwandlung Orkhund";
+
+    return (G1CP_ConstReplaceStr(strConst, originalStr, correctedStr) > 0);
```
szapp, line 9: This is good example of the comments in #361. I think these lines can be reduced to the following without loss of readability.
```D
return G1CP_ConstReplaceStr("TXT_SPELLS", "Verwandlung Orc-Hund", "Verwandlung Orkhund");
```

szapp (changes requested):
Let's discuss this one, too.

src/Ninja/G1CP/Content_G1.src
AmProsius: Agreed.

src/Ninja/G1CP/Content/Misc/constant.d
szapp: Sounds good! I would add another `e` or `em` in the end for `G1CP_ReplaceStringConstArrEle` or `1CP_ReplaceStringConstArrElem`. Thoughts?

AmProsius: I would go with `G 1CP_ReplaceStringConstArrElem` then.

src/Ninja/G1CP/Content/Symbols/string.d
```diff
@@ -241,3 +241,17 @@ func int G1CP_FindStringConstArrIdxI(var int symbId, var string needle) {
 func int G1CP_FindStringConstArrIdx(var string name, var string needle) {
     return G1CP_FindStringConstArrIdxI(MEM_GetSymbolIndex(name), needle);
 };
+
+/*
+ * Replace element in string array constant
```
szapp, line 246: We should clarify some more details in this header comment. Maybe something similar to the function above  `Find a string in a constant string array (case-sensitive) and replace it if found`.

szapp (changes requested):

src/Ninja/G1CP/Content/Symbols/string.d
szapp: And maybe: `and returns true on success`.

src/Ninja/G1CP/Content/Symbols/string.d
```diff
@@ -241,3 +241,17 @@ func int G1CP_FindStringConstArrIdxI(var int symbId, var string needle) {
...
+ */
+func int G1CP_ConstReplaceStr(var string strConst, var string originalStr, var string correctedStr) {
```
szapp: Not to get too nit-picky now, but - for consistency - the other functions of these low-level groups use `needle` and `replace` as parameter names (see [string functions](https://github.com/AmProsius/gothic-1-community-patch/blob/4ae9164a4d3a0d50eb027a09634f2410b6a34a9a/src/Ninja/G1CP/Content/Bytecode/string.d#L132)). That might spare some confusion down the line, because there is concept of a "corrected" string in this low-level function.

szapp (changes requested):
Just one last thing.

src/Ninja/G1CP/Content/Symbols/string.d
AmProsius: Nit-picky is just alright!
I completely agree.

szapp (approved):

PR #365, opened 2023-01-10
Refactor replace string constant fix

### LINKED ISSUE #43: Missing whitespace for skill point(s)

AmProsius:
### Description
Closes #43


szapp:
@AmProsius I just looked over this and realized this refactoring impacts the fix quality. Originally, the string constants were updated only if both where found as expected. In the current implementation, the fix is reported as *not applied* even if one string has been successfully replaced. This causes some inconsistencies. I think we should revise or discuss this a bit further.

AmProsius:
Then let's split the fix into two separate fixes as suggested in #43. Feels better anyway.

PR #368, opened 2023-01-13
Refactor change dialog line speaker fixes

AmProsius:
Refactor the fixes where the speaker of a dialog line is changed.

### DIFF
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix142_TemplarVIPDialog.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix142_TemplarVIPDialog.d
@@ -2,6 +2,5 @@
  * #142 Dialog lines with ambient templar interchanged pt. 1
  */
 func int G1CP_142_TemplarVIPDialog() {
-    var int funcId; funcId = G1CP_GetFuncId("Info_Tpl_8_WichtigePersonen_Info", "void|none");
-    return (G1CP_ReplaceOuInst(funcId, 0, "Info_Tpl_8_WichtigePersonen_15_02", self, other, other, self) > 0);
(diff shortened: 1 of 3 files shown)


templates/changeDialogLineSpeaker/session.d
```diff
@@ -6,5 +6,6 @@ func int G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}() {
     // ### TODO: Replace OU name of "Xxx_Ou" ###
     // ### TODO: Optional: Switch `other` and `self` ###
```
szapp, line 7: As proposed #370, I would like to pull these grouped comments inline, see other review comment on demo.

templates/changeDialogLineSpeaker/session.d
```diff
@@ -6,5 +6,6 @@ func int G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}() {
...
 
-    return G1CP_OuSwitchSpeaker("Xxx_Info", "Xxx_Ou", other, self);
+    var int funcId; funcId = G1CP_GetFuncId("Xxx_Info", "void|none");
+    return (G1CP_ReplaceOuInst(funcId, 0, "Xxx_Ou", self, other, other, self) > 0);
```
szapp, line 10:
```suggestion
    var int funcId; funcId = G1CP_GetFuncId("### TODO info function ###", "void|none");
    return (G1CP_ReplaceOuInst(funcId, 0, "### TODO OU name ###", self, other, other, self) > 0); // ### TODO Adjust ordering of self, other ###
```

szapp:
While doing some more refactoring in the future we might have a new idea for making a convenient template functions for this fix. For now, I think we already have a nice improvement on the template.

PR #369, opened 2023-01-13
Refactor change item int variable fixes

AmProsius:
Refactor the fixes where an item instance int variable is changed.

### DIFF
--- a/src/Ninja/G1CP/Content/Misc/item.d
+++ b/src/Ninja/G1CP/Content/Misc/item.d
@@ -1,0 +1,8 @@
+/*
+ * Change the value of an item's int variable. Returns true on success.
+ */
+func int G1CP_ItemChangeIntVar(var string name, var string prop, var int originalInt, var int correctedInt) {
+    var int itemId; itemId = G1CP_GetItemInstId(name);
+
+    return (G1CP_ReplaceAssignInt(itemId, 0, ConcatStrings("C_Item.", prop), 0, originalInt, correctedInt) > 0);
+}
--- a/src/Ninja/G1CP/Content/Fixes/Session/fix044_BugmeatCategory.d
+++ b/src/Ninja/G1CP/Content/Fixes/Session/fix044_BugmeatCategory.d
@@ -10,7 +10,5 @@ func int G1CP_044_BugmeatCategory() {
     ITEM_KAT_NONE = G1CP_GetIntConst("ITEM_KAT_NONE", 0, ITEM_KAT_NONE);
     ITEM_KAT_FOOD = G1CP_GetIntConst("ITEM_KAT_FOOD", 0, ITEM_KAT_FOOD);
 
-    // Replace any assignments to mainflag
-    var int itemId; itemId = G1CP_GetItemInstId("ItAt_Meatbug_01");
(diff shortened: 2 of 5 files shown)


src/Ninja/G1CP/Content/Misc/item.d
```diff
@@ -1,3 +1,12 @@
+/*
+ * Change the value of an item's int variable. Returns true on success.
+ */
+func int G1CP_ItemChangeIntVar(var string name, var string prop, var int originalInt, var int correctedInt) {
+    var int itemId; itemId = G1CP_GetItemInstId(name);
+
+    return (G1CP_ReplaceAssignInt(itemId, 0, ConcatStrings("C_Item.", prop), 0, originalInt, correctedInt) > 0);
```
szapp, line 7: I think this is really good. One thing we should consider is the the item also has integer arrays (`C_ITEM.count[5]`). In order to be able to replace, say `count[3]` specifically, we would either have to expose one more argument (as `G1CP_ReplaceAssignInt` does it) - maybe much better, parse the `prop` name to infer it:

```D
G1CP_ItemChangeIntVar("My_Item_Instance", "count[5]", oldVal, newVal);
```

For that we'd need to find an efficient way to identify and extract the number from the `prop` otherwise that might be too expensive and slow down applying the fixes.

src/Ninja/G1CP/Content/Misc/item.d
```diff
@@ -1,3 +1,12 @@
+/*
```
szapp, line 1: Let's also make this a TPL function and move it into a respective file.

src/Ninja/G1CP/Content/Misc/item.d
```diff
@@ -1,3 +1,12 @@
...
+ * Change the value of an item's int variable. Returns true on success.
+ */
+func int G1CP_ItemChangeIntVar(var string name, var string prop, var int originalInt, var int correctedInt) {
```
szapp, line 4: How about `originalInt` and `correctInt`?

szapp (commented):
We would have to discuss how to solve these comments.

src/Ninja/G1CP/Content/Misc/item.d
szapp: Or `originalVal` and `correctVal`

szapp:
The simplification on the array index (see 7e5d76e8d3e269734d5d48cc61eada6a3f27909c) causes a lot of changes all over the code base. To keep the changes traceable, I committed the low-level changes first followed by the refactoring of fixes and tests separately.

A few words on the simplification:

It's about the symbols that hold arrays, e.g. integer arrays like `C_Item.count[3]`. When checking, getting, or setting symbols, the index always had to be supplied - even for symbols that are not arrays. That resulted in a lot of function calls with lots of confusing zeros, like `G1CP_SetIntVar("InExtremoPlaying", 0, TRUE)` or looking worse with more function parameters. To simplify development and make function calls more intuitive, I moved the index into the symbol name for such getter functions. That reduces the example to a very intuitive `G1CP_SetIntVar("InExtremoPlaying", TRUE)`. Arrays would look like that: `G1CP_GetStringId("C_Item.text[1]")`.

I hope this will make our life easier.

AmProsius:
Very cool! I'll review the changes later today

PR #370, opened 2023-01-14
Refactor change VOB location fix

AmProsius:
Refactor the fixes where the location of a VOB is changed.


AmProsius:
> While I agree that the coordinates should be only written out once, I am not convinced that we need the extra wrapping function.
> 
> The abstraction is good, but we create many small functions with overlapping intent but lacking clarity in their names. The abundance of functions will make it difficult in the future to find our way through seemingly identical functions (compare G1CP_MoveVobToPos and G1CP_ChangeVobLocation).

My intent was to name the function like the fix template. But I agree that the naming may lead to confusion.

> I would propose to reduce the refactor to outsourcing the coordinates but retaining the original calling logic (using G1CP_FindVobByPosPtr and G1CP_MoveVobToPosPtr instead).

I'm okay with that. I'll revisit the fix as you suggested.

AmProsius:
The _todo_ action failed, because I edited a fix template. We can ignore that.

templates/changeVobLocation/gamesave.d
```diff
@@ -4,7 +4,7 @@
 
 // ### TODO: Replace original position of {0, 0, 0} ###
 // ### TODO: Replace corrected position of {0, 0, 0} ###
-// ### TODO: Replace callbacks of of Hlp_Is_WhateverYouSeek ###
+// ### TODO: Replace callbacks of Hlp_Is_Todo ###
```
szapp, line 7: I'd like to move this block of to-dos rather inline to where the developer has to make adjustments. That clarifies the location of the changes, one doesn't have to adjust to-dos in two places, and thanks to Botwyn's comments the locations will be nicely highlighted.

Example:
```D
const float G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}_OriginalPos[3] = {0, 0, 0}; // ### TODO Add coordinates, e.g. from spacer
```

Botwyn can handle the "TODO" with and with a following colon. I would opt for leaving it out, to reduce the overall line length in the templates.

templates/changeVobLocation/gamesave.d
```diff
@@ -28,6 +28,6 @@ func int G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}Revert() {
         return FALSE;
     };
 
-    return G1CP_ChangeVobLocation(G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}_CorrectedPos, 
-        G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}_OriginalPos, Hlp_Is_WhateverYouSeek);
-};
\ No newline at end of file
+    return G1CP_ChangeVobLocation(_@(G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}_CorrectedPos),
+        _@(G1CP_{ISSUE_NUM_PAD}_{SHORTNAME}_OriginalPos), Hlp_Is_Todo);
```
szapp, line 32: Another name for the placeholder `Hlp_Is_Todo` could be `callback` to make it clearer what it is. (I was confused what "Hlp_Is_Todo" means). A comment with "TODO" next to it (see other review comment) will make it quite convenient and clear.

src/Ninja/G1CP/Content/Misc/world.d
```diff
@@ -257,3 +257,19 @@ func int G1CP_GetWorldTime() {
 
     return hour * 100 + min;
 };
+
+/*
+ * Find VOB in the world by its exact position and move it to another position. Returns true on success.
+ *
+ * Simplifies fix template for "changeVobLocation"
+ */
+func int G1CP_ChangeVobLocation(var int originalPosPtr, var int correctedPosPtr, var int callback) {
```
szapp, line 266: I'd like to change the name of this function to relate better to the existing find and move function and to stick to the suffix of "Pos", "PosPtr", and "PosF". Both will help in usage without having to look into the implementation - once familiar with the suffixes.

I suggest `G1CP_FindAndMoveVobToPosPtr(int posPtr, int movPtr, int callback)` and moving it just below the `G1CP_MoveVobToPosF` function to group the VOB position manipulation functions.

szapp, line 266: The parameter `callback` should be of type `func` to allow supplying `NOFUNC` when no callback is desired, see the find-functions in the file above.

szapp: The function is a bit atypical compared to the others in that file, because it has a return type (which **must not** be left unpopped). If we end up making more functions maybe we can consider grouping them in another file. For now this seems good.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix0051_StonehengeCryptChest.d
```diff
@@ -1,29 +1,30 @@
 /*
  * #51 Chest in stonehenge crypt behind wall
  */
+
+/*
+ * Make the positions available to the functions below
+ */
+const float G1CP_0051_StonehengeCryptChest_OriginalPos[3] = {-35498.5625, 2331.01416, -13830.9131};
+const float G1CP_0051_StonehengeCryptChest_CorrectedPos[3] = {-35311.8100, 2281.0100, -14096.8200};
```
szapp, line 9: Since we change these anyway, "PosOld" and "PosNew" would result in shorter names (as done in #50). What do you think?

src/Ninja/G1CP/Content/Fixes/Gamesave/fix0057_CastleWallTexture.d
```diff
@@ -1,38 +1,39 @@
...
+const float G1CP_0057_CastleWallTexture_OriginalPos[3] = {-5893.91455, -664.270386, 78.9339142};
+const float G1CP_0057_CastleWallTexture_CorrectedPos[3] = {-5902.01807, -664.270386, 82.8507462};
 
 /*
  * Callback function to check the found VOBs: Check if the VOB's visual is a decal
  */
 func int G1CP_0057_CastleWallTextureCheck(var int vobPtr) {
     var zCVob vob; vob = _^(vobPtr);
-    const int zCDecal_classDef = 8861528; //0x873758
+    const int zCDecal_classDef = 8861528; // 0x873758
```
szapp, line 16: Fun fact: Although odd-looking at first, it kind of established itself to have no leading space for comments that describe memory addresses. This is not really a convention, but I like to stick to how Ikarus started it 😄

src/Ninja/G1CP/Content/Fixes/Gamesave/fix0057_CastleWallTexture.d
```diff
@@ -1,38 +1,39 @@
...
+
     return (MEM_GetClassDef(vob.visual) == zCDecal_classDef);
 };
 
+/*
+ * Apply the fix
+ */
+func int G1CP_0057_CastleWallTexture() {
+    return G1CP_ChangeVobLocation(_@(G1CP_0057_CastleWallTexture_OriginalPos),
+        _@(G1CP_0057_CastleWallTexture_CorrectedPos), G1CP_0057_CastleWallTextureCheck);
```
szapp, line 26: I still have to get used to this way of breaking lines. I agree that it looked also odd before. What do you think about breaking after and before the "(" and ")" as well in such a case? That would resemble the OTB-like bracket style used in the project.

```suggestion
    return G1CP_ChangeVobLocation(
        _@(G1CP_0057_CastleWallTexture_PosOld), _@(G1CP_0057_CastleWallTexture_PosNew),
        G1CP_0057_CastleWallTextureCheck
    );
```
I hope line wrapping doesn't make this example look too weird.

szapp (changes requested):
I blasted a lot of nit-picky comments. Read on a good day 😃 
As always, let's discuss what works best.

src/Ninja/G1CP/Content/Misc/world.d
AmProsius: Following our fix templates, I would name this `G1CP_Tpl_ChangeVobLocation()` and move it to the templates file.

szapp: That's a good suggestion. Let's go with that. When only used in the fixes anyway, the usage will be clear from code provided by the template.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix0051_StonehengeCryptChest.d
AmProsius: Okay for me. I like consistency.

src/Ninja/G1CP/Content/Fixes/Gamesave/fix0057_CastleWallTexture.d
szapp: Great compromise! I think that's how styles of other languages do it, too, right? It looks familiar from black/ruff in Python, maybe also in Prettier for js/ts?

szapp (approved):
Looks great! I will try this in-game and run the tests before merging.

szapp:
I had another idea to simplify the test of the template: During the manual test, the player was teleported to a waypoint the developer had to specify. Searching for a suitable waypoint is extra overhead and slows down development.

Instead, there is now a new function that teleports the player (or NPC) to the nearest waypoint from specified coordinates, falling back to the coordinates if none found. For the template, that means the test does not need to be adjusted at all!

...theoretically: Sometimes the nearest waypoint is inconvenient, e.g. behind a wall, and choosing a waypoint manually might sometimes be required. But where it works, it's a good start to make the template simpler.

Have a look at the changes. The refactored tests still pass. All good from my side.

